PR #3, opened 2021-01-18
Fix migration of comma-separated tracking codes (e.g. DHL)

b1u3:
In a previously used Shopware 5 setup I was using Pickware DHL Adapter to create shipments, but were not able to migrate the tracking codes created by this plugin. Therefore I fixed this issues by also migrating multiple tracking codes, which are provided as a comma-separated string.
My fix might not fit all use cases, nevertheless I want to provide my solution to you.


MalteJanz:
See ticket https://issues.shopware.com/issues/PT-12235

PR #4, opened 2021-01-26
Allow returning write results in classes implementing the WriterInterface

felixbrucker:

Currently there is no way to return the write results in a writer imlementing the `WriterInterface` interface, as its return type is `void`. This is a problem when decorating an existing Writer, like the `ProductWriter`. In this case the decorating class can not use the decorated classes `writeData()` method if it depends on the result of it, but has to re-implement the code of the `AbstractWriter` to allow working with the result of the write operation. Thus when multiple decorator of different plugins exist only one of them might be executed as the decorated instance is never used.

This PR fixes this by adding a return type to the `WriterInterface` interface. The `AbstractWriter` will now return the write result of the upsert operation and can thus be used by decorators relying on the write result.

### 3. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [ ] I have written or adjusted the documentation according to my changes
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - MIG-187 - Improves the migration of media without filename
 - MIG-188 - Improves media download stability
 - MIG-189 - Fix migration of product line items
+- Improve the extendability of the plugin
 
 # 2.2.2
 - MIG-110 - Improves the migration of media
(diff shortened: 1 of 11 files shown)


mitelg:
thanks for your contribution @felixbrucker   👍🎉💙

PR #5, opened 2022-02-18
Read document data only for matching document IDs.

kleinmann:
Without this, the migration assistant loads all order documents on every iteration. This leads to ... suboptimal runtimes and loads of errors if you limit the migrated orders in any way.


mitelg:
thanks for your contribution @kleinmann 👍🎉💙

it is already released with version 4.2.5

PR #8, opened 2023-05-05
fix: Remove superfluous column select

Wysselbie:
The additional `config.column_name` lead to not existing `column_name` key in data arrays.

Removing it solved the problem and should have no effects on the migration logic.

Tested with 5.0.0. From Shopware 5.7 to Shopware 6.5.

### DIFF
--- a/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php
+++ b/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php
@@ -81,7 +81,7 @@ abstract class AttributeReader extends AbstractReader
         $columns = $this->cleanupColumns($columns, $foreignKeys);
 
         $query = $this->connection->createQueryBuilder()
-            ->select('config.column_name, config.*')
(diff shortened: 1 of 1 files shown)


mitelg:
hey @Wysselbie 

first of all, a big apology for the long time it took us to process your PR. 

I had a look at your changes and unfortunately we could not merge it like this. 
It breaks the logic here: https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php#L141-L142
with your changes the `configuration` is `null`. Where do you have issues with the missing `column_name`? also it is not really missing, you can still find it under the `name` array key, of the result set.

mitelg:
Hey @Wysselbie 

This PR was closed due to inactivity. If this change is still important to you, feel free to create a new pull request. 

For more information about our contribution guidelines, see https://app.recruitee.com/#/offers/frontend-developer-vuejs-mwd-2

PR #10, opened 2023-07-01
PHP 8.1 compatibility

Schrank:
Since PHP 8.1, the data for a SELECT is no longer returned as PHP strings by default, but now has correct data types such as integer or float.

https://stackoverflow.com/questions/73169064/swag-migration-run-exception-in-logs-while-migrating-from-shopware5-to-shopware6

@mitelg Thanks!


Schrank:
@mitelg It seems this doesn't help 🤔  Are you sure this helps? If yes I'll test more, but for the moment I try to downgrade :-)

Schrank:
It doesn't help because it is a Magento migration, this will most likely help for Shopware 5 :D

mitelg:
thanks for your contribution @Schrank :+1: 

I am sorry to say, that this is already fixed here some lines below https://github.com/shopware/SwagMigrationAssistant/blob/master/Profile/Shopware/Gateway/Connection/ConnectionFactory.php#L73

PR #13, opened 2023-11-10
MIG-876 - feature: shipping methods with custom rules will be created…

ghost:
Although already stated within commit ... the intention of this pr is to let the assistant create shipping methods of type custom rule. I consider it a feature since shipping methods are currently not scope of current state of this extension. So its kind of a "new" functionality to users.

**Why is this patch useful?**
Because i and several others face the problem that through current restrictions migration of orders will fail due to missing shipping methods.

**What does this patch do?**
It allows creating those shipping method on migration system.

**What doesn't this patch do?**
Migrating custom rules of those shipping methods, since it can be very complex and in a lot of cases not possible to fully automate


MalteJanz:
Fixed with https://issues.shopware.com/issues/MIG-881

PR #14, opened 2024-01-08
Optimize performance for order import using the local Shopware 5 database connection

jankal:
Additionally: Add option to keep existing data and thus resume an existing MigrationRun.
Also added new Command options to set the step size for data import.


King-of-Babylon:
Hi @jankal,

thank you for your contribution, we have merged your commit :tada: 

https://github.com/shopware/SwagMigrationAssistant/commit/c60c5afff7fb6ce33c1f94fdff27113b3ff7c6e2

PR #15, opened 2024-04-02
Update DataSelectionCollection.php

warki:
Remove PHP deprecation warning:
 `Deprecated: uasort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero`


mitelg:
thanks for your contribution @warki 👍🎉💙

PR #18, opened 2024-09-08
fix: Make connection optional

aragon999:
if the assocation is not loaded it will otherwise yield a type error:

```
{"errors":[{"status":"500","code":"0","title":"Internal Server Error","detail":"Typed property SwagMigrationAssistant\\Migration\\Mapping\\SwagMigrationMappingEntity::$connection must not be accessed before initialization"}]}
```

### DIFF
--- a/src/Migration/Mapping/SwagMigrationMappingEntity.php
+++ b/src/Migration/Mapping/SwagMigrationMappingEntity.php
@@ -19,7 +19,7 @@ class SwagMigrationMappingEntity extends Entity
 
     protected string $connectionId;
 
-    protected SwagMigrationConnectionEntity $connection;
+    protected ?SwagMigrationConnectionEntity $connection = null;
 
     protected ?string $entity;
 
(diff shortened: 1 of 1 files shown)


jozsefdamokos:
@aragon999 can i ask how did you come across this error? I would like to reproduce it

aragon999:
@jozsefdamokos I have a Python script, which is performing some additional migrations, but for that I needed to fetch the original mappings of the old IDs to the new UUIDs, my request looks like this:
```python
result = sw6_client.call('search/swag-migration-mapping', data={
    "includes": {
        "swag_migration_mapping": ["oldIdentifier", "entityUuid"]
    },
    'associations': {
        'connection': {}
    },
    'filter': [
        {
            "type": "multi",
            "operator": "and",
            "queries": [
                {
                    'field': 'entity',
                    'type': 'equals',
                    'value': entity,
                },
                {
                    'field': 'oldIdentifier',
                    'type': 'equalsAny',
                    'value': ids,
                },
            ],
        },
    ],
})
```

In that case I would not need the `connection` association. But if I leave this out, the error as shown initially is thrown.

jozsefdamokos:
@aragon999 thanks for your contribution!

PR #19, opened 2024-10-06
chore: Remove remnants of psalm

aragon999:
As far as I can tell the psalm things are not needed anymore and the `bin/static-analyze.sh` script has been moved to a composer script.


jozsefdamokos:
@aragon999 thanks for the contribution!

PR #20, opened 2024-10-06
fix: Command in already running exception message

aragon999:
There is not `migration:get-status` command.


jozsefdamokos:
@aragon999 thanks for your contribution!

PR #27, opened 2025-03-03
fix: undefined array key salesChannelId exceptions in SW6->SW6

MalteJanz:
closes MIG-1099

It also does:
- PHPUnit will now emit all PHP warnings in it's output
- PHPUnit will now fail if any PHP warnings were triggered, this also fails the CI job
- Fix all PHP warnings uncovered by tests

### DIFF
--- a/tests/_fixtures/Shopware6/PageSystemConfig/02-AlreadyExists/output.php
+++ b/tests/_fixtures/Shopware6/PageSystemConfig/02-AlreadyExists/output.php
@@ -9,5 +9,5 @@ return [
     'configurationKey' => 'core.basicInformation.contactPage',
     'configurationValue' => 'b0dfd6d77eea4cbaa02b35e8562d823d',
     'salesChannelId' => '91f23b8a27c845c39ce4a2ea1187b3b9',
-    'id' => '91f23b8a27c845c39ce4a2ea1187b3b9',
+    'id' => '91f23b8a27c845c39ce4a2ea1187b3b1',
 ];
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,7 +4,8 @@
          xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd"
          bootstrap="tests/TestBootstrap.php"
          executionOrder="random"
-         cacheDirectory="var/cache/.phpunit.cache">
+         cacheDirectory="var/cache/.phpunit.cache"
+         displayDetailsOnTestsThatTriggerWarnings="true">
     <coverage/>
 
     <php>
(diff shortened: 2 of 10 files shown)


tests/_fixtures/Shopware6/PageSystemConfig/02-AlreadyExists/output.php
```diff
@@ -9,5 +9,5 @@
     'configurationKey' => 'core.basicInformation.contactPage',
     'configurationValue' => 'b0dfd6d77eea4cbaa02b35e8562d823d',
     'salesChannelId' => '91f23b8a27c845c39ce4a2ea1187b3b9',
-    'id' => '91f23b8a27c845c39ce4a2ea1187b3b9',
+    'id' => '91f23b8a27c845c39ce4a2ea1187b3b1',
```
MalteJanz, line 12: Fyi: changed this id because it was identical to the `salesChannelId` and thus might be mapped wrongly. With two different IDs for the mapping the test case should be more robust.

But unfortunately none of these tests detected the issue 🤔

MalteJanz: Nevermind, the tests actually did produce a "warning". I will investigate a bit further...

jozsefdamokos: I also had a similar issue where the test did not trigger an error just a warning, but then it would error while actually running it. I guess it's based on your php config how these are handled, but at least we could make the stricter way the default. wdyt?

larskemper (approved):

ennasus4sun (approved):

tests/_fixtures/Shopware6/PageSystemConfig/02-AlreadyExists/output.php
MalteJanz: Yes that's also what I'm thinking and will try to do. In my POV all these warnings should be visible in the PHPUnit output and actually fail the CI pipeline. Because it not only depends on your PHP configuration but also on the version, as newer PHP versions often raise hard errors for things that were warnings before and thus would fail migrations.

MalteJanz: Done.

@mitelg just wondering: platform also displays PHP warnings but doesn't fail execution. Is this intentional?. e.g. we have CI jobs that succeed and because we have so many jobs, warnings like these are easily lost, one example can be found here in platform:
https://github.com/shopware/shopware/actions/runs/13784089389/job/38548007884#step:7:67

mitelg: you have to ask @nfortier-shopware :slightly_smiling_face:

jozsefdamokos (approved):

PR #28, opened 2025-03-07
update administration asset paths

bschulzebaek:
During https://github.com/shopware/shopware/pull/7349 we found out that the asset handling is not working properly - fixing it also requires extensions to adjust their asset paths.


MalteJanz (approved):
Great discovery and thanks for fixing 🚀

larskemper (approved):

MalteJanz (approved):

PR #30, opened 2025-04-01
Add design tokens

Haberkamp:
# What

This PR starts the migration to the Design Tokens.

# Why

Design tokens make it possible that the Administration can have a dark mode.

# How

As the migration approach, we take an incremental approach. We migrate tokens in smaller badges to avoid large merge conflicts.

# Additional notes

I installed a stylelint plugins with a) auto-fixes a handful of issue and b) make people aware that they should use Design Tokens instead of hard-coded values or outdated SCSS variables.

The rules are only warnings, so they do not block the merge process.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,6 +69,9 @@
             "@npm:admin run lint-fix",
             "@npm:acceptance run lint-fix"
         ],
+        "admin:lint:scss": [
+            "@npm:admin run lint:scss"
+        ],
         "admin:unit": "@npm:jest run unit",
         "admin:unit:watch": "@npm:jest run unit-watch",
         "admin:acceptance": "@npm:acceptance run test",
--- a/composer.json
+++ b/composer.json
@@ -62,12 +62,14 @@
         "admin:lint": [
             "@npm:jest run lint",
             "@npm:admin run lint",
-            "@npm:acceptance run lint"
+            "@npm:acceptance run lint",
+            "@npm:admin run lint:scss"
         ],
         "admin:lint:fix": [
             "@npm:jest run lint-fix",
             "@npm:admin run lint-fix",
-            "@npm:acceptance run lint-fix"
+            "@npm:acceptance run lint-fix",
+            "@npm:admin run lint:scss-fix"
         ],
         "admin:lint:scss": [
             "@npm:admin run lint:scss"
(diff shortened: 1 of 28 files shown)


src/Resources/app/administration/package.json
```diff
@@ -7,7 +7,8 @@
   "license": "MIT",
   "scripts": {
     "lint": "eslint --ext .js,.ts,.vue,.html,.html.twig .",
-    "lint-fix": "npm run lint -- --fix"
+    "lint-fix": "npm run lint -- --fix",
+    "lint:css": "stylelint ./**/*.scss"
```
MalteJanz: Can you please also add `lint:css` to the general `lint` script? Because then it would also be included in our CI pipeline. The alternative would be to create an additional CI job for it.

I think something like this should work:
```
"lint": [
    "eslint --ext .js,.ts,.vue,.html,.html.twig .",
    "npm run lint:css",
],
```

MalteJanz: Also I just took a look at platform admin:
https://github.com/shopware/shopware/blob/207f83b14ef269b6e8186b6e6beb5fa76c3cbb21/src/Administration/Resources/app/administration/package.json#L34-L35

I think it would be great to stay consistent and introduce `lint:scss` + `lint:scss-fix` instead of `lint:css`

MalteJanz (changes requested):
Overall LGTM 👍 and thanks for taking this effort 💪 . Only thing I am missing is proper CI pipeline integration and optionally consistent naming of the command 🙂

src/Resources/app/administration/package.json
Haberkamp: Ok, I've added a new job.

Haberkamp: I updated the command to `lint:scss` and added a fix command for the linting.

composer.json
```diff
@@ -69,6 +69,9 @@
             "@npm:admin run lint-fix",
             "@npm:acceptance run lint-fix"
         ],
+        "admin:lint:scss": [
+            "@npm:admin run lint:scss"
+        ],
```
MalteJanz: While you have to fix the pipeline: I would suggest merging this into the general `admin:lint` composer command + also adding the fix version to `admin:lint:fix`.

I think this would be better regarding developer experience. There is just one command to lint all admin files in composer 🙂

MalteJanz (commented):
Looks like the pipeline still fails: https://github.com/shopware/SwagMigrationAssistant/actions/runs/14236613574/job/39897012689?pr=30#step:3:81

composer.json
Haberkamp: Ok, done

MalteJanz (approved):

ennasus4sun (approved):

PR #33, opened 2025-05-07
chore: prepare release 15.0.1

jozsefdamokos:


CHANGELOG_de-DE.md
```diff
@@ -1,3 +1,6 @@
+# 15.0.1
+- Verwerfungswarnungen für das Argument `tagged` in Dienstdefinitionen behoben
```
ennasus4sun: I think we should use also the english terms here... :
```suggestion
- Deprecation-Warnungen für das tagged-Argument in Service-Definitionen behoben
```

ennasus4sun (approved):

ennasus4sun (approved):

philippdinkhoff (approved):

PR #34, opened 2025-05-13
fix: connection name with hypens

### LINKED ISSUE #8013: Do a major Assistant + Magento Release to support the RC1 of 6.7
Now that the plugins are compatible in the {{trunk}}branch with shopware 6.7 we should do another release of them

### LINKED ISSUE #9155: custom-fields are no longer migrated when the connection-name contains a hyphen
### Shopware Version

6.7.0.1 (probably the same in 6.7.0.0)

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

The migration of custom fields/attributes fails if the connection name contains a string with a hyphen like "my-connection"

These errors are thrown:
```
[error] SWAG_MIGRATION_RUN_EXCEPTION
An exception occurred
Entity: order_custom_field, sourceId: 0196ae9e37017342be821ea520451f7b
Invalid custom field name: It must begin with a letter or underscore, followed by letters, numbers, or underscores.
```
This change was done in this commit: https://github.com/shopware/shopware/commit/bd8e1d0f7895e7089d8fb750a02377f6f747b6fa

We build a new technical name for the custom fields and sets, which includes the connection name: (https://github.com/shopware/SwagMigrationAssistant/blob/e520d0985342b7fce5a07458267a81bed05f9997/src/Profile/Shopware/Converter/AttributeConverter.php#L61
We could validate this before saving the values or replace the hyphen with an underscore or something else.

### Expected behaviour

Custom fields should still be migrated as before 6.7.0.1.

### How to reproduce

- insert a string with a hyphen when creating the connection to the source system
Image

jozsefdamokos:
fixes [#9155](https://github.com/shopware/shopware/issues/9155)

fixes https://github.com/shopware/shopware/issues/8013

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 15.0.2
-- Fixed data selection tab freezing due to infinite loop
+- Fixed data selection tab freezing due to infinite loop in SW6.7RC4 or later
 - Fixed connections names containing hyphens causing errors for custom field migration
 
 # 15.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 15.0.2
+- Fixed data selection tab freezing due to infinite loop
+- Fixed connections names containing hyphens causing errors for custom field migration
+
 # 15.0.1
 - Fixed deprecation warnings for `tagged` argument in services definitions
 
--- a/CHANGELOG_de-DE.md
+++ b/CHANGELOG_de-DE.md
@@ -1,5 +1,5 @@
 # 15.0.2
-- Einfrieren der Datenauswahl-Registerkarte aufgrund einer Endlosschleife behoben
+- Einfrieren des Browsers im Datenauswahl-Tab aufgrund einer Endlosschleife ab SW6.7RC4 behoben
 - Verbindungsnamen mit Bindestrichen, die Fehler bei der Migration benutzerdefinierter Felder verursachen, wurden behoben
 
 # 15.0.1
--- a/CHANGELOG_de-DE.md
+++ b/CHANGELOG_de-DE.md
@@ -1,3 +1,7 @@
+# 15.0.2
+- Einfrieren der Datenauswahl-Registerkarte aufgrund einer Endlosschleife behoben
+- Verbindungsnamen mit Bindestrichen, die Fehler bei der Migration benutzerdefinierter Felder verursachen, wurden behoben
+
 # 15.0.1
 - Deprecation-Warnungen für das tagged-Argument in Service-Definitionen behoben
 
--- a/src/Migration/Connection/Helper/ConnectionNameSanitizer.php
+++ b/src/Migration/Connection/Helper/ConnectionNameSanitizer.php
@@ -15,6 +15,7 @@ class ConnectionNameSanitizer
     /**
      * Sanitizes the connection name by removing any special characters, hyphens and spaces.
      * Only alphanumeric characters are allowed.
+     * We sometimes use the connection name in places where a unique technical name is needed and e.g. CustomFields
      */
     public static function sanitize(string $connectionName): string
     {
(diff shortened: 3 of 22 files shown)


src/Resources/app/administration/src/module/swag-migration/page/swag-migration-data-selector/index.js
```diff
@@ -38,15 +38,15 @@ Component.register('swag-migration-data-selector', {
         },
 
         fetchTableData() {
-            if (this.dataSelectionTableData.length > 0) {
-                this.$nextTick(() => {
```
jozsefdamokos, line 42: For some reason this is now running into an infinite loop causing the page to freeze and crash. I added a `setTimeout` now to wait until the ref is available. If there are better solutions let me know. I tried multiple `$nextTick()` methods but it did not work.

@shopware/product-ct-admin any ideas?

ennasus4sun (approved):

src/Resources/app/administration/src/module/swag-migration/page/swag-migration-data-selector/index.js
jleifeld: `setTimeout` is not really a solution. I think it makes more sense to debug (using the Chrome devtools) the issue of the infinite loop. Because a hotfix like this will lead to other problems in the future

src/Migration/Connection/Helper/ConnectionNameSanitizer.php
```diff
@@ -0,0 +1,23 @@
...
+class ConnectionNameSanitizer
+{
+    /**
+     * Sanitizes the connection name by removing any special characters, hyphens and spaces.
+     * Only alphanumeric characters are allowed.
+     */
+    public static function sanitize(string $connectionName): string
+    {
+        return \preg_replace('/[^a-zA-Z0-9]/', '', \str_replace([' ', '-'], '', $connectionName));
+    }
```
MalteJanz: LGTM 👍 , we could optionally mention why this constraint exists in the first place: we sometimes use the connection name in places where a unique technical name is needed and e.g. CustomFields have their own constraints on names, see
https://github.com/shopware/shopware/blob/df7ae4bf10431177298ed5bd8e32a6da0ba3c93f/src/Core/System/CustomField/CustomFieldService.php#L29

src/Resources/app/administration/src/module/swag-migration/page/swag-migration-data-selector/index.js
MalteJanz: We can take a closer look together if you like, I'm also not feeling well with the `setTimeout` workaround, even `$nextTick` wasn't that great to begin with but might have been necessary 🤷‍♂️

MalteJanz (approved):

CHANGELOG_de-DE.md
```diff
@@ -1,3 +1,7 @@
+# 15.0.2
+- Einfrieren der Datenauswahl-Registerkarte aufgrund einer Endlosschleife behoben
```
jozsefdamokos: Any suggestions?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 15.0.2
+- Fixed data selection tab freezing due to infinite loop
```
MalteJanz:
```suggestion
- Fixed data selection tab freezing due to infinite loop in SW6.7RC4 or later
```

CHANGELOG_de-DE.md
MalteJanz:
```suggestion
- Einfrieren des Browsers im Datenauswahl-Tab aufgrund einer Endlosschleife ab SW6.7RC4 behoben
```

MalteJanz (approved):

ennasus4sun (approved):

MalteJanz (approved):

PR #38, opened 2025-08-11
feat: add migration logging required fields

larskemper:
resolves [#11810](https://github.com/shopware/shopware/issues/11810)

### DIFF
--- a/src/Migration/MigrationContext.php
+++ b/src/Migration/MigrationContext.php
@@ -21,7 +21,7 @@ class MigrationContext extends Struct implements MigrationContextInterface
 
     private ProfileInterface $profile;
 
-    private SwagMigrationConnectionEntity $connection;
+    private ?SwagMigrationConnectionEntity $connection;
 
     private string $runUuid;
 
--- a/src/Migration/MigrationContext.php
+++ b/src/Migration/MigrationContext.php
@@ -21,7 +21,7 @@ class MigrationContext extends Struct implements MigrationContextInterface
 
     private ProfileInterface $profile;
 
-    private ?SwagMigrationConnectionEntity $connection;
+    private SwagMigrationConnectionEntity $connection;
 
     private string $runUuid;
 
@@ -54,8 +54,10 @@ class MigrationContext extends Struct implements MigrationContextInterface
         return $this->profile;
     }
 
-    public function getConnection(): ?SwagMigrationConnectionEntity
+    public function getConnection(): SwagMigrationConnectionEntity
     {
+        \assert($this->connection instanceof SwagMigrationConnectionEntity);
+
         return $this->connection;
     }
 
--- a/src/Profile/Shopware/Converter/CustomerConverter.php
+++ b/src/Profile/Shopware/Converter/CustomerConverter.php
@@ -99,13 +99,21 @@ abstract class CustomerConverter extends ShopwareConverter
         $this->runId = $migrationContext->getRunUuid();
         $this->migrationContext = $migrationContext;
 
+        $connection = $migrationContext->getConnection();
+        $this->connectionId = '';
+        $this->connectionName = '';
+        if ($connection !== null) {
+            $this->connectionId = $connection->getId();
+            $this->connectionName = $connection->getName();
+        }
+
         $fields = $this->checkForEmptyRequiredDataFields($data, $this->requiredDataFieldKeys);
 
         if (!empty($fields)) {
             $this->loggingService->addLogEntry(new EmptyNecessaryFieldRunLog(
                 $this->runId,
-                DefaultEntities::CUSTOMER,
-                $data['id'],
+                $connection->getProfileName(),
+                $connection->getGatewayName(),
                 \implode(',', $fields)
             ));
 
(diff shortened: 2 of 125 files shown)


src/Migration/MigrationContext.php
```diff
@@ -54,8 +54,10 @@ public function getProfile(): ProfileInterface
         return $this->profile;
     }
 
-    public function getConnection(): ?SwagMigrationConnectionEntity
+    public function getConnection(): SwagMigrationConnectionEntity
```
larskemper, line 57: @MalteJanz is there a reason why the connection object is nullable? Should we change? Would make things easier and we could delete a lot of duplicate "null checks"

MalteJanz: I agree, it would make things a lot easier to reason about and remove all these null checks all over the place.

We have to double check if there are any cases where the connection could actually be null (or why it was implemented like this in the first place).

But even when there are a few edge cases that don't have a connection. We could also build a different Context class for these that just doesn't have access to the connection 🤔

Feel free to experiment with this idea, ideally in a new issue / pr to keep things small, but as I see you already started, so just make sure this PR doesn't grow too much 😅

larskemper: I will create a new issue for that and add a `assert()` for now

larskemper: fyi: [#11883](https://github.com/shopware/shopware/issues/11883)

tests/Profile/Shopware/Gateway/Local/ProductAttributeReaderTest.php
```diff
@@ -49,12 +49,12 @@ public function testRead(): void
 
         $data = $this->productAttributeReader->read($this->migrationContext);
 
-        static::assertCount(20, $data);
-        static::assertSame('attr1', $data[0]['name']);
-        static::assertSame('text', $data[0]['type']);
+        static::assertCount(21, $data);
```
larskemper, line 52: @MalteJanz I'm not quite sure why these changed/failed now with my changes and not earlier. Comparing them to `sw55.sql`, they seem fine, but I don't fully understand the details to be certain.

tests/acceptance/tests/MigrationByUiFlow.spec.ts
```diff
@@ -14,6 +14,10 @@ test('As a shop owner I want to migrate my data from my old SW5 shop to SW6 via
     EntityCounter,
     MediaProcessObserver,
 }) => {
+    // TODO: fix & update snapshots
+    // eslint-disable-next-line playwright/no-skipped-test
+    test.skip();
```
larskemper, line 19: fyi: I skip it for now until we implemented the new logs

tests/Profile/Shopware/Gateway/Local/ProductAttributeReaderTest.php
MalteJanz: Strange, maybe this change in your MR caused this?
https://github.com/shopware/SwagMigrationAssistant/blob/1ebff3d539bd5efbea01ba9b17dd8e257f665aed/src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php#L191

On trunk it looks like this:
https://github.com/shopware/SwagMigrationAssistant/blob/f41e6abf6a6d4e5adca9698bd5b37a4cd58c4a3c/src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php#L191

src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -99,13 +99,21 @@ public function convert(
         $this->runId = $migrationContext->getRunUuid();
         $this->migrationContext = $migrationContext;
 
+        $connection = $migrationContext->getConnection();
+        $this->connectionId = '';
+        $this->connectionName = '';
+        if ($connection !== null) {
```
CR0YD, line 105: Why do you need this `if` statement here?
Doesn't return the migration conext always a connection now?
Which you then also use in the following statements.

CR0YD: Or will these cases be covered in [#11883](https://github.com/shopware/shopware/issues/11883)?

tests/Profile/Shopware/Gateway/Local/ProductAttributeReaderTest.php
larskemper: Good one, that's it. I changed it because `getLocalColumns()` was deprecated and replaced with `getReferencingColumnNames()`. For some reason, the previous method didn’t return the "first" column, so I think this behavior is correct now.

src/Profile/Shopware/Converter/CustomerConverter.php
larskemper: Exactly, this will be refactored with [#11883](https://github.com/shopware/shopware/issues/11883) and we discused it here: [ref](https://github.com/shopware/SwagMigrationAssistant/pull/38#discussion_r2272354256)

CR0YD (approved):

ennasus4sun (approved):

PR #39, opened 2025-08-15
refactor!: convert to ts

larskemper:


ennasus4sun (approved):

src/Resources/app/administration/src/module/swag-migration/extension/sw-dashboard-index/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        return {
+            context: Shopware.Context.api,
+            runExists: false,
+            loading: true,
+            run: null,
+        };
+    },
+
+    computed: {
+        migrationRunRepository(): TRepository<'swag_migration_run'> {
```
CR0YD, line 38: As you're using this generic `swag_migration_run` quite often, you could consider changing it to an enum or something similar.
But I don't know if we already do that anywhere or are always doing it way way.

CR0YD (approved):

src/Resources/app/administration/src/module/swag-migration/extension/sw-dashboard-index/index.ts
larskemper: just adopted a platform pattern here 👍

PR #40, opened 2025-08-18
refactor: add migration logging optional fields

larskemper:
resolves [#11811](https://github.com/shopware/shopware/issues/11811)


src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -552,23 +536,16 @@ protected function applyCountryTranslation(array &$country, array $data): void
...
 
         $state = ['countryId' => $newCountryId];
 
         if (!isset($oldAddressData['state_id'], $oldAddressData['country']['countryiso'], $oldAddressData['state']['shortcode'])) {
-            $this->loggingService->addLogEntry(
-                new UnknownEntityLog(
-                    $this->runId,
-                    $connection->getProfileName(),
-                    $connection->getGatewayName(),
-                    DefaultEntities::CUSTOMER,
```
ennasus4sun, line 569: just for my understanding: we want to pass this data in a later PR? (see your comments "_// TODO: add optional fields_"). Should be possible already in this one (`withEntityName()`, `withSourceData()`), true?

ennasus4sun (approved):
I like it 💙 💪🏻

src/Profile/Shopware/Converter/CustomerConverter.php
larskemper: yes, it's alreaddy possible, but I didn't wanted to make this pr bigger as it's already is and I suspect, that filling all the logs with data, to be quite a big effort. So I added some sub issues to this one

ennasus4sun: I would have at least used the ones that were already there. In my opinion, that's easier than gathering everything again in the next PR.
But that's fine for me too. Currently, you would have to collect everything again, too.

src/Migration/Logging/Log/Builder/SwagMigrationLogBuilder.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * @template T of AbstractSwagMigrationLogEntry
+     *
+     * @param class-string<T> $logClass The class name of the log entry to create
+     *
+     * @return T The created log entry instance
+     */
+    public function build(string $logClass): AbstractSwagMigrationLogEntry
+    {
+        return new $logClass(
```
jozsefdamokos: Shouldnt we check if logClass is indeed an instance of AbstractSwagMigrationLogEntry?

src/Profile/Shopware/Converter/TranslationConverter.php
```diff
@@ -676,15 +635,9 @@ protected function createConfiguratorOptionGroupTranslation(array $data): Conver
...
-            $connection = $this->migrationContext->getConnection();
-
             $this->loggingService->addLogEntry(
-                new AssociationRequiredMissingLog(
-                    $this->runId,
-                    $connection->getProfileName(),
-                    $connection->getGatewayName(),
-                    DefaultEntities::TRANSLATION
-                )
+                SwagMigrationLogBuilder::fromMigrationContext($this->migrationContext)
```
jozsefdamokos, line 639: Do these also need TODOs?

src/Migration/Logging/Log/Builder/SwagMigrationLogBuilder.php
larskemper: It's checked by phpstan, but better be safe 👍

src/Migration/Logging/Log/Builder/SwagMigrationLogBuilder.php
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * @template T of AbstractSwagMigrationLogEntry
+     *
+     * @param class-string<T> $logClass
+     *
+     * @return T
+     */
+    public function build(string $logClass): AbstractSwagMigrationLogEntry
+    {
+        $log = new $logClass(
```
jozsefdamokos, line 131: If the `$logClass` is invalid class name or the constructor arguments dont match it will throw an error. So I think we need a try ... catch here.

PR #41, opened 2025-08-19
feat: error resolution

### LINKED ISSUE #10417: Give merchants understandable error messages during migration so they can troubleshoot the most common errors on their own.
### Problem we must solve

Customers often reach out to Shopware support or developer support teams because they encounter errors during the migration and want to troubleshoot solutions. In many cases the issue appears somewhere in the source system data. The current solution involves a log file that is very abstract and doesn’t provide enough context for even technical users to solve issues independently. Customers reach out to support or even dev support, which costs these teams valuable time. 

### WHY we need to solve it

Furthermore, the constant stream of dev-related work introduces great unpredictability for the Runtime-Terror team, as many of these issues take 2-3 days to resolve. The team estimate for time spent per week on this topic is around 40%. This hinders planning for other roadmap goals. 

### Solution to the problem

The overall idea is to move the work of solving migration related issues into the hands of the merchants / users of the migration-assistant. To enable them to do this, the process of solving issues has to be as simple as possible and at the same time provide all the necessary details on issues.

For this we would like to introduce an interactive UI where users are asked to solve issues:
- individually (e.g. by specifying or correcting wrong values)
- in bulk by applying the same solution / value to all issues of the same type (if applicable, doesn't work for "unique" fields)
- (optional) by intentionally ignoring them, but being aware of all the consequences

There is likely still a small amount of issues the users can't solve themselves, e.g. technical exceptions that are caused by bugs in the migration-assistant. For these we would like to keep a migration log that contains more details for developers to troubleshoot and fix these bugs in code (specifically exception stack traces and insights about the processed data).

We came to the conclusion that a solution like this would be the best way to fix the underlying issue we have and properly reduce our support workload. Other considered solutions, e.g. of just improving the current logging solution with more context will still require a lot of technical expertise or be really annoying to fix one by one in the source system (e.g. SW5 backend).

Ideal outcomes, including proposed measurable KPIs:
- % less contact rate regarding migration-assistent
- % less dev support tickets regarding migration-assistent

### WHO benefits

- Merchants using the Migration-Assistant
- External developers / agencies using the Migration-Assistant
- Customer support for quicker troubleshooting / less requests
- Us / Developer support because less requests and time spend doing dev support

### WHERE will we do it

All changes will happen in https://github.com/shopware/SwagMigrationAssistant and related repositories

### Acceptance criteria

- [x] For the common migration case, I can either migrate successfully or get helpful error messages that I can resolve myself and migrate my data successfully afterwards
- [x] For any further migrations that have the same issues, my fixes are persisted, ~~can be adjusted~~ (see https://github.com/shopware/shopware/issues/14721 in second iteration) and are applied again
- [x] In case an error is resolved in the source system, where previously there was a fix in the target system, the source system takes priority and the old fix isn't applied again (does not override my source system data if valid)
- [x] For the uncommon migration case that ran into exceptions, I get a log file that helps developers pin point and fix the issue in the code

### Definition of Done

- [x] Fulfills all acceptance criteria defined during discovery.
- [x] Integration/E2E testing in staging is done.
- [x] All integration/E2E/unit tests passing; all critical or high-priority bugs are resolved.
- [x] Fulfills compliance, performance, security, and cloud-readiness needs.
- [x] Observability—includes monitoring, alerting, and logging; incident response handbooks updated.
- [x] Documentation—developer docs are written or updated, including information on how the feature or change is adopted and tested. If new functionality, end user documentation is written or updated.
- [x] (conditional) If the change affects the Administration, it has been tested in current versions of Firefox, Chrome, and Edge.
- [x] (conditional) If a new Admin module is created, the access control layer is provided, and permissions can be managed via the Users & Permissions module. Minimum: verify that only admins can migrate or create a ACL permission for it
- [x] (conditional) Performance testing done if the change has a high performance impact, such as affecting scalability, load balancing, or query optimization.
- [x] ~~make fixes editable or at least deletable (delete all or specific)~~ (see second iteration https://github.com/shopware/shopware/issues/14721 )
- [x] ~~make sure unnecessary data get's truncated with the "Truncate Migration" action~~ (see second iteration https://github.com/shopware/shopware/issues/15318)

related internal idea backlog item: https://shopware.atlassian.net/browse/FL-138

### LINKED ISSUE #13898: Migration: Incorrect transmission of orders with different tax rates (items)
### Shopware Version

6.7.5.0

### Affected area / extension

Platform(Default)

### Actual behaviour

Shopware 6.7.5.0
Shopware 5.7.18

Within a Shopware 5 order, there are items with different VAT rates. A correct invoice was generated within Shopware 5.

Image

1. After migrating to Shopware 6, the order in Shopware 6 is only displayed with one tax rate:

Image

2. If a cancellation invoice is generated, all tax rates are listed and the cancellation invoice is incorrectly displayed.

Image

### Expected behaviour

As with Shopware 5, the tax rates must be correctly assigned and only one tax rate per item must appear within the cancellation invoice.

### How to reproduce

.

### LINKED ISSUE #14656: Migration: Set app_id prevents migration of custom field sets
### Shopware Version

6.7.7.0 with SwagMigrationAssistant 15.0.4

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

When you try to do a migration, custom fields will always be migrated. However, when custom fields were created by an app and an app_id is defined in the table custom_field_set, the following error will occur:

```
[error] SWAG_MIGRATION_RUN_EXCEPTION
An exception occurred
Entity: custom_field_set, sourceId: 019c1f13a681727db708759ee7d58bfb
An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`mkuhlmann_sw67`.`custom_field_set`, CONSTRAINT `fk.custom_field_set.app_id` FOREIGN KEY (`app_id`) REFERENCES `app` (`id`) ON DELETE CASCADE ON UPDATE CASCADE)
```

This will also happen when the app is installed in the target shop as the app_id will most likely differ. It does not seem to impact the migration of products though, as they are migrated including the custom field content, which may be accessed when the field is restored on the target system (i.e. by installing the app)

### Expected behaviour

The custom field would be recreated or when the app exists on the target system, the different app_id would be accounted for

### How to reproduce

- Set up two shops and install the following extension in both: https://store.shopware.com/de/webar75355339825f/shop-starter-paket.html
- Create a product
- Link the system and start a migration

larskemper:
resolves shopware/shopware#10417

fixes shopware/shopware#13898
fixes shopware/shopware#14656


.github/workflows/admin.yaml
```diff
@@ -13,8 +13,8 @@ on:
   workflow_call:
```
larskemper, line 13: todo before merge: check all commit message for conventional commit and **don't squash merge** (?)

yusufttur (approved):

MalteJanz:
~~Fyi: I'm doing an interactive rebase to rename these commits to follow our conventional commit naming and force push that:~~

r 001ed934 # 12367/add entity classes (#76)
r d6af580e # 13248/change swag migration fix table (#81)
r 8a16844a # Chore: 13247/change table swag migration logging (#80)
r abde7120 # prettier

Unfortunately that didn't work out and we tried a few different approaches, all led to conflicts and differences in the files afterwards (likely because of the previous merges and conflict resolution). So we decided to stick to this history

MalteJanz (approved):

.github/workflows/admin.yaml
MalteJanz: see comment below: https://github.com/shopware/SwagMigrationAssistant/pull/41#issuecomment-3965116389

MalteJanz:
Looks like our acceptance test went red. Turns out it's not a good idea to include exception trace in our snapshot test, if changes to unrelated code can make them fail 🙈 . We should remove them before comparing the snapshot, as in the actual log they are really useful.

Screenshot 2026-02-26 at 15 22 19

jozsefdamokos (approved):

PR #43, opened 2025-08-21
refactor!: migration connection usage

larskemper:
resolves [#11883](https://github.com/shopware/shopware/issues/11883)


MalteJanz (changes requested):
Nice one 💪 . Just keep in mind that this includes breaking changes, so please document them in the `UPGRADE.md`.

Additionally it would be nice if you could open a PR on the Magento profile repository as well, so we don't have to adjust to all of these breaking changes later / at once but instead try to keep them compatible as we go 🙂

ennasus4sun (approved):
really nice! 💪🏻

larskemper:
> Additionally it would be nice if you could open a PR on the Magento profile repository as well ...

fyi: https://github.com/shopware/SwagMigrationMagento/pull/31

CR0YD (approved):

PR #47, opened 2025-08-27
fix: phpstan log array types

larskemper:
the phpstan types for arrays in logs, need to be more generic 😅

### DIFF
--- a/src/Migration/Logging/Log/Builder/AbstractSwagMigrationLogEntry.php
+++ b/src/Migration/Logging/Log/Builder/AbstractSwagMigrationLogEntry.php
@@ -18,10 +18,10 @@ abstract readonly class AbstractSwagMigrationLogEntry implements SwagMigrationLo
     final public const LOG_LEVEL_DEBUG = 'debug';
 
     /**
-     * @param array<int, array<string, mixed>>|null $sourceData
-     * @param array<int, array<string, mixed>>|null $convertedData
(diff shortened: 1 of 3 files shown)


src/Migration/Logging/Log/Builder/AbstractSwagMigrationLogEntry.php
```diff
@@ -18,10 +18,10 @@
     final public const LOG_LEVEL_DEBUG = 'debug';
 
     /**
-     * @param array<int, array<string, mixed>>|null $sourceData
-     * @param array<int, array<string, mixed>>|null $convertedData
-     * @param array<string, mixed>|null $usedMapping
-     * @param array<int, array<string, mixed>>|null $exceptionTrace
+     * @param array<mixed>|null $sourceData
```
CR0YD, line 21: Is there really no way to narrow them down even a little bit more?
This seems to kind of defeat the whole purpose of typing.
Or is `mixed` really the appropiate type here?

CR0YD (approved):

src/Migration/Logging/Log/Builder/AbstractSwagMigrationLogEntry.php
larskemper: Basically the whole code is is not types with phpstan and ignored, so that would be quite an effort. Maybe later when we get closer to finishing, but yeah of course I would also like this 👍

PR #51, opened 2025-09-04
feat: add migration validation of converted data

larskemper:
resolves [#11813](https://github.com/shopware/shopware/issues/11813)

covered validation:
- missing required fields
- unexpected fields
- validation of field value
- lookup of foreign key mapping


src/Migration/Validation/SwagMigrationValidationResult.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return SwagMigrationLogEntry[]
+     */
+    public function getLogs(): array
+    {
+        return $this->logs;
+    }
+
+    public function hasErrors(): bool
+    {
+        return \count($this->logs) !== 0;
```
jozsefdamokos, line 40: log could also be warning right? or should we name it error everywhere?

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -0,0 +1,229 @@
...
+                throw MigrationException::unexpectedNullValue($fkFieldName);
+            }
+
+            $referenceEntity = $fkField->getReferenceEntity();
+
+            if (!$referenceEntity) {
+                throw MigrationException::unexpectedNullValue($fkFieldName);
+            }
+
+            $hasMapping = $this->mappingService->hasValidMappingByEntityUuid(
```
jozsefdamokos, line 213: Does it need to have a mapping already? Or can it have a mapping set later, after this entity?

jozsefdamokos:
LGTM so far

src/Migration/Validation/SwagMigrationValidationService.php
larskemper: When I understand it correct the converters are executed in the right order, so all linked foreign entities should have been mapped before this, or am I wrong @MalteJanz?

src/Migration/Validation/SwagMigrationValidationResult.php
larskemper: yes, true. I will simply renamed it to `hasLogs()`

src/Migration/Service/MigrationDataConverter.php
```diff
@@ -95,6 +97,17 @@ private function convertData(
                 $convertStruct = $converter->convert($item, $context, $migrationContext);
                 $convertFailureFlag = empty($convertStruct->getConverted());
 
+                $validationResult = $this->validationService->validate(
```
ennasus4sun: just a thought:
would it makes sense or is it possible to introduce a new validation step/process instead handling it directly in the converter?
> read > convert > validate > write

ennasus4sun (commented):
looks good to me, too.
just one thought:

src/Migration/Service/MigrationDataConverter.php
larskemper: Definitely possible, the only difference is that the validation would then run after all converters have finished. The ticket says "after each converter".
@MalteJanz  is there a particular reason for that, or would you be okay with us introducing it as a separate step instead?

src/Migration/Validation/SwagMigrationValidationService.php
MalteJanz: In theory yes, otherwise we would also run into issues while writing all the data. e.g. all `DataSelection`'s have a position in which they are ordered (e.g. -100 for basicSettings) and the order of the returned `DataSet`'s matters too:
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Profile/Shopware/DataSelection/BasicSettingsDataSelection.php#L41-L60

So all readers, converters, writers should follow that execution order.

But just having an entry in the mapping table of course doesn't mean the foreign entity will actually be there at the time this entity is going to be written (it could have it's own errors preventing it from being written before). Ideally these will not be an issue if the user fixes all issues and everything gets written correctly.

There is also a catch to this approach: Every time we call `$mappingService->getOrCreateMapping(...)` instead of `$mappingService->getMapping(...)` it will create an entry in the mapping table if none exists and there are a few places where we do the former while migrating a parent entity, for example here:
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Profile/Shopware/Converter/ProductConverter.php#L711-L716

Means if that product data with all its foreign entities fails to be converted or written it also means that specific manufacturer might not be there (if not successful in another product). And if a later stage relies on that manufacturer to be there (e.g. `ManufacturerAttributeConverter` + `Writer`) it will also fail to write


TLDR: I guess this approach should be fine for now, at least while users fix all open issues and don't ignore anything

src/Migration/Service/MigrationDataConverter.php
MalteJanz: A different step would also work but I'm leaning a little against that if you don't have good reasons that would suggest otherwise. My reasoning why it should happen in the same step:

- The converted data is already there, otherwise you would write it to the DB and later fetch it again, which takes additional time.
- Introducing another step also introduces additional overhead, you need to keep in mind that this process / steps run on potentially millions of rows (is not very uncommon in practice). Additionally our `limit` + `offset` pagination isn't great and slows down at larger offsets.

In hindsight I'm also not a fan of splitting the reading (+converting) and writing steps but with our change to fix the issues before write it at least has a purpose to stay that way. I don't know why it was build initially like that 🤷 .

ennasus4sun: hey, it was more or less a feeling when reading the code: mixed purposes and responsibilties.
source-data (converting issues) ✅ and validation against our Definitions.
but yes, I understand the benefit of "data is already there" 👍🏻 

Another question that comes to mind when thinking about this: could it be, that we duplicate errors for the same issue and displaying them in the UI?
example:
- converter logs "missing email"
- validator logs also "missing required field"
🤔 

if yes: maybe we can add an "unduplication"....

MalteJanz: About the duplication, good point. Yes most converters have some very rudimentary checks for "required fields". We should try to remove all of them and instead use this new, proper validation based on the actual entity definitions.

ennasus4sun (approved):

DennisGarding (approved):

PR #52, opened 2025-09-05
fix: phpstan for feature branch

larskemper:


src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php
```diff
@@ -134,11 +135,21 @@ private function getAttributeConfiguration(string $table, MigrationContextInterf
             $translationStrPos = (int) \mb_strrpos($translation['name'], '_');
             $field = \mb_substr($translation['name'], $translationStrPos + 1);
 
-            if (!isset($attributeConfiguration[$column]['translations'][$field])) {
-                $attributeConfiguration[$column]['translations'][$field] = [];
+            $col = &$attributeConfiguration[$column];
```
jozsefdamokos, line 138: I find this harder to understand than how it was before, but thats just personal preference

jozsefdamokos (approved):

src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php
larskemper: agree, but for some reason that's the only solution phpstan likes 😅

DennisGarding (approved):

PR #53, opened 2025-09-09
feat: add error resolution ui

larskemper:
resolves [#13015](https://github.com/shopware/shopware/issues/13015)

related https://github.com/shopware/shopware/pull/13587

## Summary

Introduces the new ui to resolve user-fixable migration errors grouped by code/entity/field.

==> [screen design](https://www.figma.com/design/V0y1816pt26PeSYYORSxNy/Migration?node-id=1246-13293)
==> [cloud version](https://ci-20251126-1310-dhawan-clicky-platform.swstage.store/admin): settings > migration (please do not confirm "continue" 😄)

## What changed

> [!NOTE]
> Tests will follow in a seperate pr to make this not bigger as it already is (pr: #90)

#### `src/Controller/HistoryController`
 
New Endpoints:
 
 - GET `api/_action/migration/get-log-groups`: Main endpoint to fetch grouped logs of the current migration run to resolve errors. It groups the logs by error code, entity name and field name. Also it gives the option to paginate, sort (by & direction) and filter the results by error code, status (resolved / unresolved), entity, and field name. Uses the new `LogGroupingService` for fetching the logs.

 - POST `api/_action/migration/get-all-log-ids`: For the error resolution modal I need a endpoint that gives me all Ids of logs groups by code, entity name and field name to select all logs of this group for bulck editing.

#### `src/Migration/History/LogGroupingService`

New Service:

- groups logs by code, entity name, field name for given runId and level
- supports filters: error code, entity name, field name, status (resolved / unresolved)
- supports sorting: by count, code, entity name, field name, profile name, gateway name createdAt
- supports pagination, respecting filters and sorting
- computes the `fixCount` for the ui where we display how many of the grouped errors have been resolved (e.g. `12 / 500`)
- computes the `levelCounts` for error type error/warning/info (e.g. `Errors (12), Warning (30) ...`)

#### New components
 
- `swag-migration-error-resolution-step`: Entry point for resolving migration issues. Lists grouped log items, offers filters and sorting, opens the resolution modal

- `swag-migration-error-resolution-modal`: Bulk resolution modal for selected group of errors. Lets the user choose a target log, input a fix value and create a fix

- `swag-migration-error-resolution-details-modal`: Shows the user the raw data of the logs convertedData and sourceData if available

- `swag-migration-error-resolution-log-filter`: Filter panel for the main table. Provides code, enity, field and status select.

- `swag-migration-error-resolution-field`: Field wrapper that inspects the entity medadata to render the correct input type (subcomponents)

  - `swag-migration-error-resolution-field-relation`: input for relation fields (to-one / to-many). User can either select one entity (id) or multiple

  - `swag-migration-error-resolution-field-scalar`: input for scalar field (stirng / number / boolean / dates / json)

  - `swag-migration-error-resolution-field-unhandeled`: fallback for unsuppoorted or unknown field types (e.g. if an unexpexted field got recognized and we have no metadata in the definitions for the field or the type of field is unsupported)

### New Service:

`src/module/swag-migration/service/swag-migration-error-resolution.service.ts`:

Business logic for the error resolution with following responsibilities:

- resolves admin route for an entity (getEntityLink) to create option for a fix
- analyses the entity definitions to extract:
  - type of scalar values
  - type and entity of associations
  - required fields
  - extraction of id fields (to-one and to-many ids)
  - entity field validation
- field utilities:
  - finds corresponding association fields in the definition
  - picks "highest priority" property of an entity for displaying meaningful information
- table generation: builds columns for the error resolution modal by following schema: 1. status, 2. field that has error, ... followed by prioritized scalar fields


src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -28,6 +28,7 @@ export const MIGRATION_API_SERVICE = 'migrationApiService';
 export const MIGRATION_STEP = {
```
larskemper, line 28: change target (should not be merged into `feature/migration-logging-refactor`)!

DennisGarding (dismissed):

MalteJanz (commented):
Just had a quick look while checking it out. Overall looks already quite close to our end goal 💪 .

But it also contains quite a few smaller bugs and isn't really usable right now (e.g. I couldn't test the "swag-migration-error-resolution-fields" because i got js errors about `undefined` objects while accessing `path`.

I guess we should wait for the backend prototype or add more dummy data to test these more in depth.

DennisGarding (approved):

ennasus4sun (approved):
as talked about in the daily

- something seems to be wrong in counting fixed entities:<br />despite the fact that all rows have been assigned a fixed value <br />
Bildschirmfoto vom 2025-11-17 09-01-52

- we could talk about how we handle entities which havent the "correct" (unexpected) field-names
Bildschirmfoto vom 2025-11-17 09-24-24

larskemper:
feedback todos:
- [x] make detail modal rows clickable (if possible): [#13587](https://github.com/shopware/shopware/pull/13587)
- [x] remove support call to action in unfixable error banner
- [x] refactor continue modal
- [x] add log table empty state
- [x] preselction of table pages does not work for select all
- [x] fix filter log popover outside click (no fix as it seems)
- [x] step indicator: increase margin between items & add horizontal divider (small viewports)
- [x] truncate log text in detail modal
- [x] add filter count badge to log filter button
- [x] fix banner darkmode
- [x] fix stepper tokens
- [x] stepper should not be sticky (no easy fix)
- [x] fix fixed/total query
- [x] code column refactor: add name column, hide code column, use code as fallback for name column

src/Controller/HistoryController.php
```diff
@@ -25,11 +26,13 @@
...
-    public function __construct(private readonly HistoryServiceInterface $historyService)
-    {
+    public function __construct(
+        private readonly HistoryServiceInterface $historyService,
+        private readonly LogGroupingServiceInterface $logGroupingService,
+    ) {
     }
 
-    #[Route(path: '/api/migration/get-grouped-logs-of-run', name: 'api.admin.migration.get-grouped-logs-of-run', methods: ['GET'], defaults: ['_acl' => ['swag_migration.viewer']])]
+    #[Route(path: '/api/_action/migration/get-grouped-logs-of-run', name: 'api.admin.migration.get-grouped-logs-of-run', methods: ['GET'], defaults: ['_acl' => ['swag_migration.viewer']])]
```
MalteJanz, line 35: This isn't an "action" it is just a `GET` request.

src/Controller/HistoryController.php
```diff
@@ -105,4 +108,105 @@ public function isMediaProcessing(): JsonResponse
 
         return new JsonResponse($result);
     }
+
+    #[Route(
+        path: '/api/_action/migration/get-log-groups',
```
MalteJanz, line 113: Because this is also only a `GET` request I would remove the `_action` from the path.

But I see, looks like we aren't consistent with this pattern everywhere 🙈.

@mitelg is this even something we still do: no `_action` in path if it is only a `GET` request? Because just looking around a bit almost all admin api endpoints have this `_action` in their path, even when they only allow `GET` 🤔

If we don't really have a convention here, I guess it would be better sticking with `_action` everywhere, so we at least stay consistent 🤷

src/Controller/HistoryController.php
```diff
@@ -105,4 +108,105 @@ public function isMediaProcessing(): JsonResponse
...
+
+        if (!\is_string($sortDirection) || !\in_array(\strtoupper($sortDirection), ['ASC', 'DESC'], true)) {
+            $sortDirection = 'DESC';
+        }
+
+        $result = $this->logGroupingService->getGroupedLogsByCodeAndEntity(
+            $runId,
+            $level,
+            (int) $pageParam,
+            (int) $limitParam,
```
MalteJanz: ~~I think there should be a better fallback than just casting to `int`, when these query params aren't available~~

Nevermind, you already throw a exception in these cases. But are these params really required?

src/Controller/StatusController.php
MalteJanz, line 380: Just curious and mostly unrelated to this PR: Is the user ever able to get out of this state again? The `isReset` flag stays active forever right? Because we want to see e.g. in dev support that someone reset their migration before.

src/Migration/Gateway/GatewayRegistry.php
```diff
@@ -25,8 +24,10 @@ public function __construct(private readonly iterable $gateways)
     /**
      * @return GatewayInterface[]
      */
-    public function getGateways(ProfileInterface $profile): array
+    public function getGateways(MigrationContextInterface $migrationContext): array
     {
+        $profile = $migrationContext->getProfile();
```
MalteJanz, line 29: Why is this change necessary? It looks odd to me that you now have to construct a context, just to call this method, while before having a profile was enough and it still doesn't use any other information from the context 🤔

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -28,7 +28,7 @@ export const MIGRATION_API_SERVICE = 'migrationApiService';
...
     IDLE: 'idle',
     FETCHING: 'fetching',
-    APPLY_FIXES: 'apply-fixes',
+    ERROR_RESOLUTION: 'apply-fixes',
```
MalteJanz, line 31: Why not staying consistent with the backend here? I would stick to `APPLY_FIXES` or whats the reason this was changed?

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -420,31 +425,153 @@ export default class MigrationApiService extends ApiService {
...
+            status: 'resolved' | 'unresolved' | null;
+            entity: string | null;
+            field: string | null;
+        },
+        additionalHeaders: AdditionalHeaders = {},
+    ): Promise<{
+        total: number;
+        items: Array<{ code: string; entityName: string | null; fieldName: string | null; count: number }>;
+        levelCounts: { error: number; warning: number; info: number };
+    }> {
```
MalteJanz, line 486: Would be nice if we extract these types into their own interfaces instead of only specifying them inline

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -420,31 +425,153 @@ export default class MigrationApiService extends ApiService {
...
+    }
+
+    async getAllLogIds(
+        runId: string,
+        code: string,
+        entityName: string,
+        fieldName: string,
+        connectionId?: string,
+        additionalHeaders: AdditionalHeaders = {},
+    ): Promise<{ ids: string[] }> {
```
MalteJanz, line 573: how many "ids" does this return? no pagination?

MalteJanz (changes requested):
Just had a quick look, will continue reviewing in more detail this afternoon, but already wanted to share the first comments 🙂

src/Controller/HistoryController.php
larskemper: until now I thought it specifies the type of api = the "action"-api, not what it does (i mean we already define the http method)

src/Controller/HistoryController.php
larskemper: I could try to ser- /deserialize the criteria object, but I don't really know if this is better. The functionality itself is needed

src/Controller/StatusController.php
larskemper: nope, I changed that cause it isn't documented, currently it will be resetted and used as indicator for the reset. But maybe it would also be enough for this purpose to log that a reset happened?

src/Migration/Gateway/GatewayRegistry.php
larskemper: this was an missed refactor from a commit before in the early state of the error-resolution implementation. I would have to look into it more again, but I guess we needed information from the context in some cases and generalized it therefore. It failed as I tried to create a cloud version, so that's why it is in this pr

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: actually the backend isn't consistent with the (quite early) implementated ui, the screen design and the name that we found for this feature. So I would rather change it in the backend

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: it's just used here as a return type and never used again, but can't hurt 👍

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: maybe we could rename this to something more specific, but in this case I actually need all ids that match the given filters. It is used to provide a "select all" feature in the detail modal, so I need all ids to check their select state

src/Controller/StatusController.php
MalteJanz: I don't think a log entry is good for this, it's likely to go unnoticed in either the normal shopware log or the migration log. I think this DB flag was a little better, but also can't say for sure, e.g. if the user uninstalled and reinstalled the plugin, it would be false again.

Maybe we also don't need that information anymore, but we could improve the confusions around "connections" in the assistant a bit. Because in practice I don't think someone every needs more than one connection, and often when someone did use two (e.g. using api and local gateway) they point to the same shop, which is completely wrong and causes issues while migrating the same data twice. The same would also apply if they migrated once, deleted their connection and recreated another to the same source system...

So maybe there is a better solution to prevent users from running into that issue 🤔

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: Also fine to me, I would only like to see consistency in the naming here and across our implementation for this "migration step". Even if the UI wording / snippet for this step is different, which is fine, at least the implementation should be consistent 🙂

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: I think we need another approach for that use case, because I don't think the UI should deal with 2 million product id's in case two million products have the same field missing for example 🙈 .

One idea from my side would be a "streaming" approach:
- show a loading spinner in the UI
- fetch one batch of ids at a time, persisting the fix for these ids
- do that until all ids that match the given criteria are processed, then stop the loading spinner / whatever, notify the user

I'm happy to talk further about this, but we should always keep in mind that there are customers with massive amounts of data and our solution should be able to handle that without running into issues, ideally also as performant as possible.

src/Controller/StatusController.php
larskemper: let's create a ticket for that 😅

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: Will create a ticket for checking/adjusting naming consistency in general 👍

src/Controller/HistoryController.php
MalteJanz: What I mean is at least for `$limitParam` I would expect a reasonable default like `100` without the need to pass it always. A bit similar for `$pageParam` like `1`. So it's more consistent with our normal repository API's which also have reasonable defaults as criteria.

And going the full way with a criteria object also sounds wrong, because certain things like aggregations aren't support by this API.

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: yeah true, but would be interesting to know how many ids it would take for this to become really inefficient 😅.

should I adjust it in this pr? I feel like we could move faster (and start manual testing) if we create a refactoring ticket for this

src/Controller/HistoryController.php
larskemper: now i get it, yes sure, will add

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: No just create a ticket, I would also prefer getting this merged as soon as possible, because it's already large enough

src/Controller/HistoryController.php
MalteJanz: To be totally honest, I have no idea where this convention came from, can't find anything in our dev docs or ADRs about it, and only vaguely remember one time, where I think @mitelg mentioned something about it to me, but maybe I remembered it wrong 🤷 .

But yes it's kind of duplicate with the HTTP method if it's only about being a "action" that is being performed and there are even cases where the same path can be used as either `GET` or `POST`, which is totally fine.

I guess now it's to late to adjust all these API paths, so I think staying consistent is the better option here. But I'm still curious if anybody else remembers what this was about? 🙈

src/Migration/History/LogGroupingService.php
```diff
@@ -0,0 +1,423 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+use SwagMigrationAssistant\Exception\MigrationException;
+use SwagMigrationAssistant\Migration\Run\SwagMigrationRunCollection;
+use SwagMigrationAssistant\Migration\Run\SwagMigrationRunEntity;
+
+#[Package('fundamentals@after-sales')]
+class LogGroupingService implements LogGroupingServiceInterface
```
MalteJanz, line 36: Can we mark this service `@internal`, or at the very least the constructor? I would prefer the former to have less public api surface here, unless there is a good reason for making it "public"

src/Migration/History/LogGroupingService.php
```diff
@@ -0,0 +1,423 @@
...
+        string $sortBy,
+        string $sortDirection,
+        ?string $filterCode,
+        ?string $filterStatus,
+        ?string $filterEntity,
+        ?string $filterField,
+        Context $context,
+    ): array {
+        $run = $this->getMigrationRunForLogs($runUuid, $context);
+        $connectionIdBytes = Uuid::fromHexToBytes($run->getConnectionId() ?? '');
```
MalteJanz: Is the `$run` only used for fetching the connection id here? In that case I would suggest doing just that in a method using plain SQL, removing the `$runRepo` dependency from this class and making it more uniform.

src/Migration/History/LogGroupingService.php
```diff
@@ -0,0 +1,423 @@
...
+        $sql = "
+            SELECT LOWER(HEX(l.id)) as id
+            FROM swag_migration_logging l
+            {$join}
+            WHERE l.run_id = :runId
+                AND l.code = :code
+                AND l.entity_name = :entityName
+                AND l.field_name = :fieldName
+                AND l.user_fixable = 1
+                AND f.id IS NULL
```
MalteJanz, line 199: Similar to the other endpoint, this is an unrestricted query, potentially returning lots of ids. We should benchmark this service and see if it actually becomes an issue

src/Migration/History/LogGroupingService.php
```diff
@@ -0,0 +1,423 @@
...
+    private function buildOrderByClause(string $sortBy, string $sortDirection): string
+    {
+        $columnMap = [
+            'count' => 'count',
+            'code' => 'l.code',
+            'entityName' => 'l.entity_name',
+            'fieldName' => 'l.field_name',
+            'profileName' => 'l.profile_name',
+            'gatewayName' => 'l.gateway_name',
+            'createdAt' => 'l.code',
```
MalteJanz: this line looks wrong, should it be `l.auto_increment` or something?

src/Migration/History/LogGroupingService.php
```diff
@@ -0,0 +1,423 @@
...
+        ];
+
+        $orderColumn = $columnMap[$sortBy] ?? 'count';
+        $direction = \in_array(\strtoupper($sortDirection), ['ASC', 'DESC'], true) ?
+            \strtoupper($sortDirection) : 'DESC';
+
+        return "ORDER BY {$orderColumn} {$direction}, l.code ASC, l.entity_name ASC, l.field_name ASC";
```
MalteJanz: I mean yes, you ensure only certain values are allowed in these places but it still screams SQL injection on first look🤯.

I know this is safe and the only possible way here, and you also did a good job in the other places passing params around. Just wanted to highlight that this is very dangerous code and it should be audited carefully. Maybe even add a comment above why this is safe (as it only allows predetermined values).

And a small nitpick: I don't think all these small methods with part of a bigger SQL statement makes it easier to read. I would much rather prefer one huge SQL query that I can easily verify that nothing bad is passed in by string concatenation ✌️

MalteJanz (changes requested):
Still not finished, didn't find that much time this afternoon 🙈 , will continue tomorrow ✌️

src/Migration/History/LogGroupingService.php
larskemper: this is the implementation of the already mentioned endpoint, so will be covered by the refactor ticket

src/Migration/History/LogGroupingService.php
larskemper: good one, actually can be removed

src/Migration/Gateway/GatewayRegistry.php
MalteJanz: Alright, just saw that `getGateway` below also takes in the context, so staying consistent here makes sense 👍

src/Controller/StatusController.php
larskemper: fyi: https://github.com/shopware/shopware/issues/13670, ok with that?

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: fyi: https://github.com/shopware/shopware/issues/13671

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: fyi: https://github.com/shopware/shopware/issues/13672, but I guess we should talk about this to find a good solution 👍

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    methods: {
+        prepareData(data: Record<string, never> | null | undefined): string {
+            if (!data) {
+                return '';
+            }
+
+            const string = JSON.stringify(data, null, 2);
+
+            if (string.trim() === '{}') {
+                return null;
```
MalteJanz, line 45: returning `null` here looks wrong, if the return type expects a `string`. Above an empty string is returned. I'm wondering if this file is checked correctly by typescript? 🤔

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    methods: {
+        prepareData(data: Record<string, never> | null | undefined): string {
+            if (!data) {
+                return '';
+            }
+
+            const string = JSON.stringify(data, null, 2);
+
+            if (string.trim() === '{}') {
```
MalteJanz: `string` can be `undefined` if the json stringify didn't work

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/swag-migration-error-resolution-details-modal.html.twig
```diff
@@ -0,0 +1,30 @@
+{% block swag_migration_error_resolution_details_modal %}
+    <sw-modal
+        class="swag-migration-error-resolution-details-modal"
+        variant="large"
+        :title="modalTitle"
+        @modal-close="() => $emit('close-error-resolution-details-modal')"
```
MalteJanz: Does the emitted event need to be this long? Also I think we follow a certain convention for event names with noun + verb 🤔 

```suggestion
        @modal-close="() => $emit('modal-close')"
```

MalteJanz (changes requested):

src/Migration/History/LogGroupingService.php
larskemper: I refactored the service and made it a little simpler by using compact sql. Also made the validation more visible and documented the security concerns for future editors

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/index.ts
larskemper: added null as return type. "checked correctly by typescript" the way we use ts at shopware: not really 😅 we would have to enable strict mode, but nothing is typed to that extend

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/swag-migration-error-resolution-details-modal.html.twig
larskemper: we have nested modals here, so I thought it wouldn't hurt being more specific, but for consistency reasons your're right

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/index.ts
MalteJanz: I'm wondering if we can at least enable `strictNullChecks` see:
https://www.typescriptlang.org/tsconfig/#strictNullChecks

So these things would get detected:
https://www.typescriptlang.org/play/?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false#code/FAMwrgdgxgLglgewgAgBYFMA2mEAoCUAXMgM4wBOcEA5sgN7DJPJRJmnIC8yA5Btgh4BuYI2bl0MMORQQw2Ec2QB6ZRKkzSIgL7AgA

And I think `null` / `undefined` safety is really important and can catch a lot of bugs early

Also the core administration seems to use `strict` mode, so whats holding us back to do the same in our plugin? 🤔 
https://github.com/shopware/shopware/blob/9881ccf31101ca565fa380097e509325c829d81f/src/Administration/Resources/app/administration/tsconfig.json#L21

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-field/swag-migration-error-resolution-field-unhandled/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                    this.updateFieldValue(parsedValue);
+                }
+            },
+            immediate: true,
+        },
+    },
+
+    methods: {
+        cleanJsonString(jsonString: string): string {
+            return jsonString.replace(/,(\s*[}\]])/g, '$1').trim();
```
MalteJanz: A doc comment would be nice here, what does the regex replace do, why is it needed?

Looks to me like it's removing the last `,` comma before an array / object gets closed. Why is this needed? The json specification denies trailing commas, so why do we need support for them here / this extra magic?

Also this magic can make mistakes:
```json
{
    "value": "My emjoi of the day ,} one eye with a mouth"
}
```
which is completely valid json and is likely still going to be stripped by this to:
```json
{
    "value": "My emjoi of the day } one eye with a mouth"
}
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-log-filter/index.ts
```diff
@@ -0,0 +1,302 @@
...
+
+            const uniqueValues = aggregation.buckets.map((bucket) => bucket.key);
+
+            this.searchResults = {
+                ...this.searchResults,
+                [type]: this.buildResultsMap(type, uniqueValues),
+            };
+        },
+
+        async loadInitialOptions(type: keyof LogFilterValue) {
```
MalteJanz: where is the difference between this method and calling `fetchSearchResults` with empty `searchTerm`? Looks like duplicate code

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-log-filter/index.ts
```diff
@@ -0,0 +1,302 @@
...
+
+        async fetchSearchResults(searchTerm: string, type: keyof LogFilterValue) {
+            const field = fieldMap[type];
+
+            const aggregationName = `${type}Aggregation`;
+
+            const criteria = new Criteria(1, 1)
+                .addAggregation(Criteria.terms(aggregationName, field, 25, null, null))
+                .addFilter(Criteria.equals('runId', this.runId))
+                .addFilter(Criteria.equals('userFixable', 1));
```
MalteJanz, line 154: If I see it correctly you are only interested in the aggregation result. I think you can also specify multiple (not nested) aggregations in a single repository search call, means you query all three `code`, `entity` and `field` at the same time instead of calling this method three times. It would also simplify the code a bit.

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-field/swag-migration-error-resolution-field-unhandled/index.ts
larskemper: the json that gets "cleaned" here is user input, I thought removing these commas would be a nice help for this common mistake. And I don't think that ",}" => "}" will be a real problem for any one 😅, but we can also remove it

MalteJanz: Yes, I also don't expect someone to have strings like this but I would still prefer no magic here and instead getting a normal error message when the json validation / parsing failed, it's better than getting a surprise that the value is slightly different than what you specified

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-details-modal/index.ts
larskemper: we would have to explore this, maybe not now/here 😅

MalteJanz: Sure, maybe another ticket 😅 🙈

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -0,0 +1,488 @@
...
+        },
+
+        async fetchEntityIdsFromMissingLogs(missingLogIds: string[]): Promise<string[]> {
+            const criteria = new Criteria(1, missingLogIds.length)
+                .addFilter(Criteria.equals('runId', this.runId))
+                .addFilter(Criteria.equals('code', this.selectedLog.code))
+                .addFilter(Criteria.equals('entityName', this.selectedLog.entityName))
+                .addFilter(Criteria.equals('fieldName', this.selectedLog.fieldName))
+                .addIncludes({ swag_migration_logging: ['entityId'] })
+                .setIds(missingLogIds);
```
MalteJanz, line 241: Are all these filters even needed when you search for specific ids anyway? Or do the passed in ids don't fulfill this criteria yet?

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,708 @@
...
+    [DATA_TYPES.JSON_LIST]: FIELD_COMPONENT_TYPES.EDITOR,
+    [DATA_TYPES.JSON_OBJECT]: FIELD_COMPONENT_TYPES.EDITOR,
+} as const;
+
+/**
+ * @private
+ * list of fields prioritized for sorting purposes, to determined most meaningful fields first.
+ */
+export const PRIORITY_FIELDS = [
+    'id',
```
MalteJanz: small nitpick: I would move id further down, as we humans can't really read out much of UUIDs and can identifier data much better by descriptive names like `name`, `title`, `label`, `technicalName`, `firstName`, ...

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,708 @@
...
+ */
+export default class SwagMigrationErrorResolutionService {
+    /**
+     * translates a migration error code into a human-readable message.
+     * if no translation is found, returns the original code.
+     */
+    translateErrorCode(code: string): string {
+        const translationKey = `${MIGRATION_ERROR_TRANSLATION_SNIPPET_PREFIX}.${code}`;
+
+        return Shopware.Snippet.tc(translationKey) ?? code;
```
MalteJanz: `tc` / `t` will return the snippet key if it isn't found. You can check if a snippet key exists with `te`:
https://vue-i18n.intlify.dev/api/injection.html#te-key-locale

additional nitpick: `tc` is basically deprecated (just points to `t` inside shopware since last major), we should only use `t` for new code. I forgot to mention that in all other places. I guess at one point we have to change all occurrences anyways, so feel free to ignore this one more time. background: https://vue-i18n.intlify.dev/guide/migration/breaking10#deprecate-tc-and-tc-for-legacy-api-mode

MalteJanz (changes requested):
Still not quite at the end 🙈 To be honest, there is quite a lot of frontend code in this PR and I'm wondering if we really need that much or can simplify stuff further.

I know some of it is needed, because we are also building a complex UI here, but there is also quite some code related to fetching and transforming data around.

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-log-filter/index.ts
```diff
@@ -0,0 +1,278 @@
...
+                return;
+            }
+
+            this.loading = true;
+
+            try {
+                await Promise.all([
+                    this.fetchSearchResults(null, 'code'),
+                    this.fetchSearchResults(null, 'entity'),
+                    this.fetchSearchResults(null, 'field'),
```
MalteJanz: I still think this is a code smell. I see at least two places where these aggregations are fetched via individual requests like this instead of together in one request. See my previous comment:
https://github.com/shopware/SwagMigrationAssistant/pull/53/files#r2559695656

Also I would suggest keeping it stupidly simple:
- just fetch all three aggregations once the component gets mounted, then just use them
- I wouldn't worry about filtering these options down further if a certain search term is inserted, let the user still select any `code` / `entity` / `field` that is available in the overall logs for filtering
- then you also don't have to worry about what happens if the user selected a certain field and then entered a search term (where there are technically no results for the field), the user will just see the result that nothing can be found and can remove / adjust the field or term. The filtering doesn't have to be smart about it in my point of view.

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+
+            if (routeName) {
+                return {
+                    name: routeName,
+                };
+            }
+        }
+
+        if (entityName.endsWith('_translation')) {
+            const baseEntityName = entityName.slice(0, -12);
```
MalteJanz: nitpick
```suggestion
        const translationSuffix = '_translation';
        if (entityName.endsWith(translationSuffix)) {
            const baseEntityName = entityName.slice(0, -translationSuffix.length);
```

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+        if (!entityName) {
+            return createEmptyEntityFields();
+        }
+
+        if (!Shopware.EntityDefinition.has(entityName)) {
+            return createEmptyEntityFields();
+        }
+
+        const definition = Shopware.EntityDefinition.get(entityName);
+        const fields = createEmptyEntityFields();
```
MalteJanz: nitpick
```suggestion
        const fields = createEmptyEntityFields();
        if (!Shopware.EntityDefinition.has(entityName)) {
            return fields;
        }

        const definition = Shopware.EntityDefinition.get(entityName);
        
```

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+        }
+
+        return null;
+    }
+
+    /**
+     * gets the entity field definition for a specific field.
+     */
+    getEntityField(entityName: string | null | undefined, fieldName: string | null | undefined): Property | null {
+        if (!fieldName || UNHANDLED_FIELD_NAMES.includes(fieldName as (typeof UNHANDLED_FIELD_NAMES)[number])) {
```
MalteJanz: Nitpick: This type wrangling makes it quite hard to read.

Suggestion A: remove `as const` from the `UNHANDLED_FIELD_NAMES` declaration so it is just a normal string array and `includes` accepts any string.

Suggestion B: cast the array here instead of the needle (because you don't know what string value the needle actually is, so why trick typescript in thinking that way):

```suggestion
        if (!fieldName || (UNHANDLED_FIELD_NAMES as readonly string[]).includes(fieldName)) {
```

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+                (property as Property & { localField?: string }).localField === fieldName
+            ) {
+                associationField = property;
+            }
+        });
+
+        // fallback: try to infer association name from field name
+        // example: "productVersionId" -> "product"
+        if (!associationField && fieldName.endsWith('VersionId') && fieldName !== 'versionId') {
+            const inferredName = fieldName.slice(0, -9);
```
MalteJanz: nitpick: see above, magic number `-9` instead of using string length that should be stripped

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+                return;
+            }
+
+            if (
+                property.type === DATA_TYPES.ASSOCIATION &&
+                (property as Property & { localField?: string }).localField === fieldName
+            ) {
+                associationField = property;
+            }
+        });
```
MalteJanz: Instead of iterating through all properties, I would try something like the following:

```ts
const associationField = Object.keys(schema.properties).find((propertyName) => {
    const property = schema.properties[propertyName];

            if (
                property.type === DATA_TYPES.ASSOCIATION &&
                (property as Property & { localField?: string }).localField === fieldName
            ) {
                return true;
            }

    return false
});
```
*That would only give you the `propertyName` though 🙈

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+            return true;
+        }
+
+        const entityField = this.getEntityField(entityName, fieldName);
+
+        if (!entityField) {
+            return true;
+        }
+
+        if (UNHANDLED_FIELD_TYPES.includes(entityField.type as (typeof UNHANDLED_FIELD_TYPES)[number])) {
```
MalteJanz: see above

```suggestion
        if ((UNHANDLED_FIELD_TYPES as readonly string[]).includes(entityField.type)) {
```

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+        }
+
+        // field type is not supported
+        return this.getFieldType(entityName, fieldName) === null;
+    }
+
+    /**
+     * determines if a field is a scalar field or should be treated as a relation field.
+     */
+    isScalarField(entityName: string | null | undefined, fieldName: string | null | undefined): boolean {
```
MalteJanz, line 396: we already have methods like this in the EntityDefinition (or `schema` as you called it sometimes here):
https://github.com/shopware/shopware/blob/6c1bc650aa4ba945ffd607cc50a27e5a9ae7b57b/src/Administration/Resources/app/administration/src/core/data/entity-definition.data.ts#L170-L172 

So are these custom implementations really needed?

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+        // id fields with corresponding association fields are treated as relation fields
+        const correspondingAssociation = this.findCorrespondingAssociationField(entityName, fieldName);
+
+        return !(entityField.type === DATA_TYPES.UUID && correspondingAssociation);
+    }
+
+    /**
+     * checks if a field is a "to many" association (one_to_many or many_to_many).
+     */
+    isToManyAssociationField(entityName: string | null | undefined, fieldName: string | null | undefined): boolean {
```
MalteJanz, line 416: see above:
https://github.com/shopware/shopware/blob/6c1bc650aa4ba945ffd607cc50a27e5a9ae7b57b/src/Administration/Resources/app/administration/src/core/data/entity-definition.data.ts#L186

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+        return this.getEntityField(entityName, fieldName);
+    }
+
+    /**
+     * determines the field type for rendering (either component type or relation type).
+     */
+    getFieldType(entityName: string | null | undefined, fieldName: string | null | undefined): string | null {
+        const entityField = this.getEntityField(entityName, fieldName);
+
+        if (!entityField || UNHANDLED_FIELD_TYPES.includes(entityField.type as (typeof UNHANDLED_FIELD_TYPES)[number])) {
```
MalteJanz: see above

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
```diff
@@ -0,0 +1,712 @@
...
+            if (priorityA !== undefined) {
+                return -1;
+            }
+
+            // only field B has a defined priority
+            if (priorityB !== undefined) {
+                return 1;
+            }
+
+            return 0;
```
MalteJanz, line 501: you could still sort by field names here instead of leaving their order unspecified if no priority was found for both of them

```suggestion
           if (a < b) return -1;
           if (a > b) return 1;
           return 0;
```

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -174,6 +174,15 @@
       }
     },
     "index": {
+      "warningModal": {
+        "title": "Dateninkonsistenz erkannt",
+        "cancelButton": "Zurück",
+        "confirmButton": "Trotzdem fortfahren",
+        "message": {
+          "currency": "Eine Migration mit <b>unterschiedlichen Standardwährungen</b> kann zu korrupten Datensätzen führen und unvorhersehbare Nebenwirkungen verursachen.",
+          "language": "Eine Migration mit <b>unterschiedlichen Standardsprachen</b> kann zu korrupten Datensätzen führen und unvorhersehbare Nebenwirkungen verursachen."
```
MalteJanz: Please try to avoid using HTML in snippets. I would keep it simple here and not make the text bold, but if you really want to, I would give this a read and use that approach:
https://vue-i18n.intlify.dev/guide/advanced/component.html

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -589,6 +599,163 @@
...
+          "SWAG_MIGRATION__SHOPWARE_UNSUPPORTED_NUMBER_RANGE_TYPE": "Nummernkreis-Typ nicht unterstützt",
+          "SWAG_MIGRATION__SHOPWARE_UNSUPPORTED_SHIPPING_CALCULATION_TYPE": "Versandkosten-Berechnungstyp nicht unterstützt",
+          "SWAG_MIGRATION__SHOPWARE_UNSUPPORTED_SHIPPING_PRICE": "Faktor-Preisberechnung bei Versandkosten nicht unterstützt",
+          "SWAG_MIGRATION__SHOPWARE_UNSUPPORTED_DOCUMENT_TYPE": "Dokumenttyp nicht unterstützt",
+          "SWAG_MIGRATION__SHOPWARE_UNSUPPORTED_MEDIA_DEFAULT_FOLDER": "Medien-Standardordner nicht unterstützt"
+        },
+        "errors": {
+          "downloadLogsFailed": "Fehler beim Herunterladen der Log-Datei.",
+          "continueMigrationFailed": "Fehler beim Fortsetzen der Migration.",
+          "fetchLogsFailed": "Fehler beim Abrufen der Log-Datei.",
```
MalteJanz:
```suggestion
          "fetchLogsFailed": "Fehler beim Abrufen der Log-Daten.",
```

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -589,6 +599,163 @@
...
+          "noLogsSelected": "Bitte wählen Sie mindestens einen Log-Eintrag zur Behebung aus.",
+          "noEntityIdsFound": "Keine gültigen Entitäts-IDs in den ausgewählten Log-Einträgen gefunden.",
+          "fetchFilterDataFailed": "Das Abrufen der Filterdaten ist fehlgeschlagen.",
+          "fetchExistingFixesFailed": "Das Abrufen vorhandener Korrekturen ist fehlgeschlagen.",
+          "invalidJsonInput": "Diese Eingabe ist kein gültiges JSON."
+        },
+        "step": {
+          "continue-modal": {
+            "title": "Fortfahren",
+            "text-unfixable": "There are {count} unfixable errors that are still unresolved. If you continue without solving the remaining errors, follow-up errors could appear.",
```
MalteJanz: english in german file?

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -589,6 +599,163 @@
...
+            "text-fixable": "There are {count} groups of fixable errors that are still unresolved. If you continue without solving the remaining errors, follow-up errors could appear."
```
MalteJanz: see above

MalteJanz: also why does this say `groups of errors` and the other just mentions `errors` 🤔

src/Resources/app/administration/src/module/swag-migration/snippet/en.json
```diff
@@ -444,10 +449,163 @@
...
+          "noEntityIdsFound": "No valid entity IDs found in the selected log entries.",
+          "fetchFilterDataFailed": "Failed fetching filter data.",
+          "fetchExistingFixesFailed": "Failed fetching existing fixes.",
+          "invalidJsonInput": "This input is not valid JSON."
+        },
+        "step": {
+          "continue-modal": {
+            "title": "Continue",
+            "text-unfixable": "There are {count} unfixable errors that are still unresolved. If you continue without solving the remaining errors, follow-up errors could appear.",
+            "text-fixable": "There are {count} groups of fixable errors that are still unresolved. If you continue without solving the remaining errors, follow-up errors could appear."
```
MalteJanz, line 509: see above the `groups of` feels of compared to the snippet right above it

MalteJanz (approved):
Finally done with my code review, sorry that it took so long 😅 ✌️ .

And great job, this PR definitely wasn't a small task 💪 . Let's try to get it finally merged now, iterating further from that 🙂

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
larskemper: the difference is that the entity definition treats uuids as a scalar value, but we want to use them as a relation. I will try to find a different name 😅

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
larskemper: cause the banner of unfixable errors mentions "errors" and the table of fixable errors displays "groups" of errors. Is ux approved 😅

src/Resources/app/administration/src/module/swag-migration/snippet/en.json
larskemper: same

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
larskemper: Object.values() instead of Object.keys() then it works 👍

src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts
MalteJanz: Ah that makes sense, still unfortunate that this implementation is so much more complicated than the core version 😅

PR #54, opened 2025-09-09
feat: delay step transition until all media is processed

### LINKED ISSUE #10822: Migration isn't finished when it says so, include media download in processing logic
### Technical TODO

Currently the migration process is in the finished state, even when there are still media files being processed. It would be great to prevent this situation and only say that the migration is completed / finished when it is actually done with everything.

Current pain points

The acceptance test can’t assert that everything is done after it says migration completed, which leads to flaky behaviour without a workaround

Customers with many media files might assume everything is done and then find that media files are broken without something in the log file. This may result in early support cases that would be solved if the customer waits a little longer (could be up to hours if there are many media files)

There is no progress indicator how far the media download has progressed

Goals of this ticket

Only transition into the finished state, if all media processing is done. At least have a loading spinner for this waiting phase.

(optional) report the actual media download progress in the corresponding media download step instead of how many messages are submitted

Ideas for implementation

Progress could be tracked by polling the swag_migration_media_files table for unprocessed and unfailed media files

The process media messages themself could report / check the overall progress at the end and transition the migration step / submit a new migration message to continue. But extra care must be taken that these messages aren’t lost / return early because of an exception

…

jozsefdamokos:
fixes: [#10822](https://github.com/shopware/shopware/issues/10822)

### DIFF
--- a/src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
+++ b/src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
@@ -10,6 +10,8 @@ namespace SwagMigrationAssistant\Migration\MessageQueue\Handler;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
 use Shopware\Core\Framework\Log\Package;
 use SwagMigrationAssistant\Exception\MigrationException;
 use SwagMigrationAssistant\Exception\NoConnectionFoundException;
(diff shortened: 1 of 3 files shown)


jozsefdamokos:
@MalteJanz do you see any issues or concerns with the approach here so far?

MalteJanz:
@jozsefdamokos Idk, you basically have to keep the following in mind with the current / old implementation:
- all reading / writing steps are processed sequentially by a single message in the message queue that get's inserted again and again
  - this works reliably 
- the only exception to this is the process media messages, we insert all of them (in chunks during the main migration messages) into the queue
  - this has the benefit that if your environment runs multiple workers on the message queue things are processed in parallel and thus should finish faster in practice than sequentially
  - But it also has drawbacks of added complexity and that you need to keep race conditions (data races) in mind which can be hard to verify and also reproduce if they occassionally happen.
  - This didn't really matter before this because all that work was kind of truly parallel and we didn't need to wait on it to finish

So the possible approaches I see here are either:

1. Simplify the media processing / download to be executed sequentially as well (as its already the case with all other migration workloads)
2. Build a reliable and data race free way to wait for all of the media processing to be finished and that only one migration process message continues afterwards (not potentially duplicate ones)

src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
```diff
@@ -101,6 +110,39 @@ public function __invoke(ProcessMediaMessage $message): void
...
+        $progress->setProgress($progress->getProgress() + \count($message->getMediaFileIds()));
+
+        $this->migrationRunRepo->update([[
+            'id' => $message->getRunId(),
+            'progress' => $progress->jsonSerialize(),
+        ]], $context);
+
+        if ($this->isAllMediaProcessed($context)) {
+            $this->runTransitionService->transitionToRunStep($migrationContext->getRunUuid(), MigrationStep::CLEANUP);
+            $this->messageBus->dispatch(new MigrationProcessMessage($context, $migrationContext->getRunUuid()));
```
MalteJanz, line 126: Just as an example of a potential data race: the two last media processing messages are handled at the same time in parallel. Both finish processing their chunk at the same time, setting the processed flag in the `swag_migration_media_file` table. After that both run this if check, which for both returns that no media files are left.

In the end we get two `MigrationProcessMessage` in the queue running the cleanup step and continuing the migration in parallel.

PR #57, opened 2025-09-18
feat: checksum and reset via mq

larskemper:
resolves [#10820](https://github.com/shopware/shopware/issues/10820)


src/Migration/MessageQueue/Handler/Processor/AbortingProcessor.php
```diff
@@ -58,7 +58,7 @@ public function process(
         MigrationProgress $progress,
     ): void {
         $connection = $migrationContext->getConnection();
-        $this->runService->cleanupMappingChecksums($connection->getId(), $context);
+        $this->runService->startCleanupMappingChecksums($connection->getId(), $context);
```
larskemper: @MalteJanz does the checksum cleanup have to be finished before aborting the migration?

MalteJanz: It should be part of the "aborting" step and wait until it is done (including showing progress) before transitioning to the next step

larskemper:
@MalteJanz / @shopware/st-runtime-terror  , what are your thoughts on the current state?

src/Resources/app/administration/src/module/swag-migration/component/card/swag-migration-shop-information/index.ts
```diff
@@ -233,6 +254,52 @@ export default Shopware.Component.wrapComponentConfig({
...
+        },
+
+        registerPolling() {
+            this.unregisterPolling();
+
+            this.migrationStore.setIsResettingChecksum(true);
+            this.pollingIntervalId = setInterval(this.resetCheckSumRoller, MIGRATION_RESET_CHECKSUM_POLLING_INTERVAL);
+        },
+
+        async resetCheckSumRoller() {
```
ennasus4sun: typo?
```suggestion
        async resetCheckSumPoller() {
```

ennasus4sun (commented):
Looks good to me so far.

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
```diff
@@ -0,0 +1,208 @@
...
+
+        $finalProgress = new MigrationProgress(
+            0,
+            0,
+            $progress?->getDataSets() ?? new ProgressDataSetCollection(),
+            $message->getEntity() ?? DefaultEntities::RULE,
+            $progress?->getCurrentEntityProgress() ?? 0
+        );
+
+        $completion = new AbortCompletionMessage(
```
jozsefdamokos: The name of this is a bit confusing to me. I think `AbortMessage` should be enough

src/Migration/Run/RunService.php
```diff
@@ -272,6 +255,24 @@ public function assignThemeToSalesChannel(string $runUuid, Context $context): vo
         $this->loggingService->saveLogging($context);
     }
 
+    /**
+     * @return list<mixed>
+     */
+    public function getMigrationMessengerStats(Context $context): array
```
jozsefdamokos: I think we better set some kind of flag on the run entity instead of directly checking the messages table for a status. Because this way we are tied to the database implementation of the messenger component

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
larskemper: true, but `AbortMessage` would suggest that you abort something, but it's actually the completion of aborting 😄

src/Migration/MessageQueue/Handler/Processor/AbortingProcessor.php
```diff
@@ -57,12 +55,12 @@ public function process(
...
-        $progress->setIsAborted(true);
-        $this->updateProgress($migrationContext->getRunUuid(), $progress, $context);
-        $this->bus->dispatch(new MigrationProcessMessage($context, $migrationContext->getRunUuid()));
+        $this->bus->dispatch(new ResetChecksumMessage(
+            $migrationContext->getConnection()->getId(),
+            $context,
+            true,
+            $run->getId(),
+            $progress->getCurrentEntity(),
+        ));
```
MalteJanz, line 66: I'm trying to understand the concept behind this. E.g. the migration process previously worked through a series of steps, see

https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Migration/Run/MigrationStep.php#L13-L42

and

https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Migration/Run/RunServiceInterface.php#L17-L27

Couldn't the "reset checksums" workload be just another step after the `ABORTING` step? Then you wouldn't need a custom message + handler + logic to properly continue the rest of the migration process afterwards. Same for the swag_migration_data table reset at the start / end of the migration (could also be its own step, which I guess is already the case now with the cleanup step right?).

The only open question then would be how can the "checksum reset" or "cleanup / finish migration" (which deletes all migration data, including the mappings, triggered in the UI) things be triggered outside of a migration run. The latter is obvious as it's not part of the migration process anyways going with a new message + handler and some way to check its progress in the admin is the way to go.

The former could also be an option for a migration run, e.g. checkbox to remove checksums while starting a new migration run and that step then needs to run before the others 🤔 . But the tricky thing then would be that this step could execute two times during a migration if it was aborted and transitioning to the proper next step needs to be handled correctly

larskemper: the problem is that checksum reset has two distinct use cases:
- standalone operation - user clicks "reset checksums" in ui.
- part of abort flow - during migration abort.

so I created a unified handler to work with both cases, to avoid different implemenations `ResetChecksumHandler`. for these cases this means a flow like this:
- standalone: reset -> ResetChecksumMessage dispatched -> ResetChecksumHandler -> finished
- migration abort: abort -> MigrationProcessHandler -> AbortingProcessor -> ResetChecksumMessage dispatched -> ResetChecksumHandler -> AbortCompletionMessage dispatched -> AbortCompletionHandler -> [continue migration flow] ....

this way I didn't had to change the migration flow itself that much

MalteJanz: TLDR: if there isn't a cleaner way we should at least document the migration flow and what message queue handlers are called when
https://github.com/shopware/shopware/issues/13016

larskemper: got rid of the abort step and moved the logic to the reset step

src/Core/Migration/Migration1759000000AddIsResettingChecksumsToSetting.php
```diff
@@ -0,0 +1,41 @@
...
+
+        if (isset($columns[self::COLUMN])) {
+            return;
+        }
+
+        $connection->executeStatement(\sprintf(
+            'ALTER TABLE %s ADD COLUMN %s TINYINT(1) NOT NULL DEFAULT 0',
+            self::TABLE,
+            self::COLUMN
+        ));
```
MalteJanz: Good practice would be using our https://github.com/shopware/shopware/blob/82692c9aa7f8e9349f67fe267b566ea04e55570a/src/Core/Framework/Migration/AddColumnTrait.php#L14 , which is already in scope (part of `MigrationStep` parent class)

src/Migration/MessageQueue/Handler/Processor/CleanUpProcessor.php
```diff
@@ -59,22 +60,64 @@ public function process(
...
         }
 
-        $this->updateProgress($migrationContext->getRunUuid(), $progress, $context);
-        $this->bus->dispatch(new MigrationProcessMessage($context, $migrationContext->getRunUuid()));
+        $query = $this->connection->createQueryBuilder()
+            ->delete('swag_migration_data')
+            ->where('id IN (:ids)')
+            ->setParameter('ids', $ids, ArrayParameterType::BINARY);
+
+        return (int) $query->executeStatement();
```
MalteJanz: Is this the fastest way to delete? Isn't it possible to execute a delete SQL statement with a limit as well? Like for what do you need the ids first? 🤔

src/Migration/MessageQueue/Handler/CleanupMigrationHandler.php
```diff
@@ -38,16 +38,25 @@ public function __invoke(CleanupMigrationMessage $message): void
...
         $nextStep = $currentStep + 1;
+
         if (isset($tablesToReset[$nextStep])) {
             $nextMessage = new CleanupMigrationMessage($tablesToReset[$nextStep]);
             $this->bus->dispatch($nextMessage);
         }
-        $this->connection->executeStatement('DELETE FROM ' . $tablesToReset[$currentStep] . ';');
+
+        $this->connection->executeStatement(
+            'DELETE FROM ' . $tablesToReset[$currentStep] . ';'
```
MalteJanz: This isn't chunking the delete and might still take multiple seconds on millions of records. We should work with limit and multiple messages as well here, otherwise a single message could still run into a timeout 🙈

src/Migration/MessageQueue/Handler/CleanupMigrationHandler.php
MalteJanz: Just a thought, maybe a better suited name for this handler and message would be something like `WipeMigrationHandler`, because we also have the `CleanUpProcessor` in each migration run.

This is essentially wiping all data related to the migration-assistant. Another word could be `Truncate` 🤔

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+
+        $query = $this->connection->createQueryBuilder()
+            ->select('m.id')
+            ->from('swag_migration_mapping', 'm')
+            ->where('m.checksum IS NOT NULL')
+            ->andWhere('m.connection_id = :connectionId')
+            ->setParameter('connectionId', Uuid::fromHexToBytes($connectionId))
+            ->setMaxResults(self::BATCH_SIZE);
```
MalteJanz: Same here, is it really necessary to select the ids to update first?

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
```diff
@@ -0,0 +1,228 @@
...
+
+        if (!$message->isResettingAll()) {
+            $query->innerJoin(
+                'm',
+                'swag_migration_data',
+                'd',
+                'd.mapping_uuid = m.id AND d.written = 0'
+            );
+        }
```
MalteJanz: In what cases is this needed / called with `isResettingAll = false`? I guess we can remove it entirely 🤔

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
```diff
@@ -0,0 +1,228 @@
...
+            $progress?->getDataSets() ?? new ProgressDataSetCollection(),
+            $message->getEntity() ?? DefaultEntities::RULE,
+            $progress?->getCurrentEntityProgress() ?? 0
+        );
+        $finalProgress->setIsAborted(true);
+
+        $this->migrationRunRepo->upsert([
+            [
+                'id' => $runId,
+                'progress' => $finalProgress,
```
MalteJanz: In the other places you execute the update like this:

```suggestion
                'progress' => $finalProgress->jsonSerialize(),
```

src/Migration/Run/RunService.php
```diff
@@ -177,30 +171,23 @@ public function abortMigration(Context $context): void
...
+
+        if ($connection === null) {
+            throw MigrationException::noConnectionFound();
         }
 
         $this->dbalConnection->executeStatement(
-            $sql,
-            [$connectionUuid],
-            [ParameterType::STRING]
+            'UPDATE swag_migration_general_setting SET `is_resetting_checksums` = 1;'
```
MalteJanz: shouldn't there be a check if it is already resetting check sums before dispatching another message as well? E.g. doing a select query before this update or checking if this update did affect a row (and adding a where clause to only update on =0)

src/Migration/Run/RunService.php
MalteJanz: see above

src/Migration/Run/RunServiceInterface.php
```diff
@@ -34,7 +34,9 @@ public function getRunStatus(Context $context): MigrationState;
      */
     public function abortMigration(Context $context): void;
 
-    public function cleanupMappingChecksums(string $connectionUuid, Context $context): void;
```
MalteJanz, line 37: might be a good idea to already add this breaking changes to our `UPGRADE.md` file, so we don't forget it. otherwise we have to do it all in the ticket github.com/shopware/shopware/issues/12625

src/Migration/Run/RunService.php
MalteJanz: Also starting a new migration run should be prevented while it's resetting checksums on the backend as well 🙂

src/Core/Migration/Migration1759000000AddIsResettingChecksumsToSetting.php
larskemper: wow, why don't I know this util 😆

src/Migration/MessageQueue/Handler/Processor/CleanUpProcessor.php
larskemper: i need to know how many entities are left to exit / redispatch

src/Migration/MessageQueue/Handler/CleanupMigrationHandler.php
larskemper: good one, didn't checked, just formatted 😅

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
larskemper: need to check how many entities are left

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
larskemper: I adopted this behavior from before, did't really find out why this is useful, so i will delete it

src/Migration/Run/RunService.php
larskemper: yep true, It's preventing / used in the admin right know, but adpoting it in the backend, makes sense 😄

DennisGarding (approved):

src/Migration/MessageQueue/Handler/Processor/CleanUpProcessor.php
MalteJanz: But you already get a number from the affected rows that were deleted and you return that here, so where is the difference?

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
MalteJanz: See above, you could also always execute update / delete statements and when the affected row count reaches zero, you are done.

src/Migration/MessageQueue/Handler/ResetChecksumHandler.php
MalteJanz: I think before it was an optional parameter defaulting to `true`. I don't know where it was called with `false` if anywhere at all. Now I'm unsure if it is called with `false` at all anymore, I think you always dispatch the message with `true` 🤔 

Edit: just doubled checked on `trunk`, it was never called explicitly with `false` and always defaulted to `true`. the parameter was even missing in the interface... So feel free to go ahead and remove this dead code

MalteJanz (approved):

PR #63, opened 2025-09-24
feat: apply fixes prototype

DennisGarding:
closes: [#12365](https://github.com/shopware/shopware/issues/12365)

Implement apply migration fix before write the converted data to DB

Answered Questions:

1. Get it implemented in an actual data migration, e.g. you mentioned the AbstractWriter as a place to put this new logic, try to get it working there

**Works now: see src/Migration/Service/MigrationDataWriter.php line 106 - 107**

2. Intentionally break some data in your source system (e.g. remove the email of one customer in SW5). Then migrate them to SW6. You should find a log entry about it and the customer shouldn't be migrated. Then manually add a fix to swag_migration_fixes and migrate again. Is the customer then migrated properly?

**Works fine and the customer is migrated.**

3. After 2. fix the SW5 customer in the source system (assign him a different email than used in the fix). Then migrate again. Takes the new source data or the old fix priority?

**Currently, there is no functionality to prioritize the source data. The fix is always applied. There is a Ticket to implement this!** See Ticket: https://github.com/shopware/shopware/issues/11819


jozsefdamokos:
@DennisGarding I dont really understand how this should work. Can you give an example of how this would be called?

src/Migration/Writer/MigrationFix/MigrationFix.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace SwagMigrationAssistant\Migration\Writer\MigrationFix;
+
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Exception\MigrationException;
+
+#[Package('after-sales')]
+class MigrationFix
+{
+    private const PATH_SEPERATOR = '.';
```
ennasus4sun:
```suggestion
    private const PATH_SEPARATOR = '.';
```

DennisGarding:
@jozsefdamokos 

We can apply the fixes before we write. Just an example. We need to figure out the best position for this, because we need the connectionId

image

ennasus4sun (approved):

src/Migration/Writer/MigrationFix/MigrationFix.php
```diff
@@ -0,0 +1,63 @@
...
+    public function apply(array &$item): void
+    {
+        $pathArray = explode(self::PATH_SEPARATOR, $this->path);
+
+        $temp = &$item;
+        foreach ($pathArray as $key) {
+            $temp = &$temp[$key];
+        }
+
+        $temp = \json_decode($this->value, true, 512, \JSON_THROW_ON_ERROR);
```
MalteJanz: the json_decode can also already happen in `fromDatabaseQuery` right? So the apply method is as simple / fast as possible

src/Migration/Writer/MigrationFix/MigrationFix.php
```diff
@@ -0,0 +1,63 @@
...
+
+#[Package('after-sales')]
+class MigrationFix
+{
+    private const PATH_SEPARATOR = '.';
+
+    public function __construct(
+        public readonly string $id,
+        public readonly string $connectionId,
+        public readonly string $mainMappingId,
```
MalteJanz: why are all these needed? I only see `$value` and `$path` being used to apply the fix

src/Migration/Writer/MigrationFix/MigrationFixMapping.php
```diff
@@ -0,0 +1,74 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace SwagMigrationAssistant\Migration\Writer\MigrationFix;
+
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Exception\MigrationException;
+
+#[Package('after-sales')]
+class MigrationFixMapping
```
MalteJanz: What purpose has this class?

src/Migration/Writer/MigrationFix/MigrationFixApplier.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @param array<string> $ids
+     *
+     * @return array<MigrationFixMapping>
+     */
+    private function getMapping(array $ids, string $connectionId): array
+    {
+        $result = $this->connection->createQueryBuilder()
+            ->select('id', 'id', 'connection_id', 'entity', 'old_identifier', 'entity_uuid', 'entity_value', 'checksum', 'additional_data')
```
MalteJanz: why are all these fields needed?

src/Migration/Writer/MigrationFix/MigrationFixApplier.php
```diff
@@ -0,0 +1,119 @@
...
+            ->from('swag_migration_mapping')
+            ->where('entity_uuid IN (:ids)')
+            ->andWhere('connection_id = :connectionId')
+            ->setParameter('ids', $ids, ArrayParameterType::STRING)
+            ->setParameter('connectionId', $connectionId)
+            ->executeQuery()
+            ->fetchAllAssociativeIndexed();
+
+        $mappingIds = array_keys($result);
+        $fixes = $this->getFixes($mappingIds, $connectionId);
```
MalteJanz: instead of retrieving the fixes after the mappings, why not selecting the fixes by an inner join and then you don't need any data about the mapping at all, except matching the mainMappingId + entity_uuid

src/Migration/Writer/MigrationFix/MigrationFixApplier.php
```diff
@@ -0,0 +1,119 @@
...
+            $mappingWithFixes->applyFixes($item);
+        }
+    }
+
+    /**
+     * @param array<MigrationFixMapping> $mappings
+     */
+    private function getMappingWithFixes(string $id, array &$mappings): ?MigrationFixMapping
+    {
+        foreach ($mappings as $index => $fixMapping) {
```
MalteJanz: why looping over all these to find the right one? Just return an associative array where the keys are the ids of the data passed in

tests/unit/Migration/Writer/MigrationFix/MigrationFixTest.php
```diff
@@ -0,0 +1,160 @@
...
+                    'doNotTouch' => 'untouchedValue',
+                    'the' => [
+                        'doNotTouch' => 'untouchedValue',
+                        'value' => [
+                            'doNotTouch' => 'untouchedValue',
+                            'which' => [
+                                'doNotTouch' => 'untouchedValue',
+                                'needs' => [
+                                    'doNotTouch' => 'untouchedValue',
+                                    'to' => [
```
MalteJanz, line 45: If you remove this `to` level with all it's children. Will the fix still apply correctly and create new nested fields? That could also be another test case

MalteJanz (changes requested):
Open questions from my side that I think this prototype should answer:

1. Get it implemented in an actual data migration, e.g. you mentioned the `AbstractWriter` as a place to put this new logic, try to get it working there
2. Intentionally break some data in your source system (e.g. remove the email of one customer in SW5). Then migrate them to SW6. You should find a log entry about it and the customer shouldn't be migrated. Then manually add a fix to `swag_migration_fixes` and migrate again. Is the customer then migrated properly?
3. After 2. fix the SW5 customer in the source system (assign him a different email than used in the fix). Then migrate again. Takes the new source data or the old fix priority?

src/Migration/Writer/MigrationFix/MigrationFixMapping.php
DennisGarding: Each mapping can have multiple fixes for a different path

src/Migration/Writer/MigrationFix/MigrationFix.php
DennisGarding: At the first step i just created a class with all informations

src/Migration/Writer/MigrationFix/MigrationFixApplier.php
DennisGarding: At the first step i just created a class with all informations

tests/unit/Migration/Writer/MigrationFix/MigrationFixTest.php
DennisGarding: Should work. I add a test for this

src/Migration/Writer/MigrationFix/MigrationFix.php
DennisGarding: Like this, we have a consistent type of the value property.  I think it doesnt matter where the jsonDecode is called, because it is short before the writing process.

src/Migration/Writer/MigrationFix/MigrationFixMapping.php
MalteJanz: Got that, but do you need to have all this mapping information just to apply the fixes? Ideally you only need a list of `MigrationFix` that applies for each corresponding `id` of data

src/Migration/Writer/MigrationFix/MigrationFix.php
MalteJanz: Got it, was thinking a bit too far about "applying the same fix to all affected entities" story, but that will be handled differently as there needs to be a backend endpoint that just inserts all the fixes into the DB...

So yeah each fix is only applied once and then it doesn't matter where this decode happens 👍

src/Migration/Writer/MigrationFix/MigrationFixMapping.php
DennisGarding: Got that to :smile:  . I am currently on it to make it smaller.

MalteJanz:
would be great if you could add a proper PR description that documents your insights and decisions from this prototype and also answers these questions 🙂 :
https://github.com/shopware/SwagMigrationAssistant/pull/63#pullrequestreview-3334083142

phpstan.neon.dist
```diff
@@ -59,7 +59,6 @@ rules:
     - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\Deprecation\DeprecatedMethodsThrowDeprecationRule
     - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\Tests\CoversAttributeRule
     - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\Tests\MockingSimpleObjectsNotAllowedRule
-    - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\FinalClassRule
```
MalteJanz: Why was this removed?

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
```diff
@@ -150,12 +150,7 @@ public function testConvertWithInvalidEmail(): void
             $this->migrationContext
         );
 
-        static::assertNull($convertResult->getConverted());
-
-        $logs = $this->loggingService->getLoggingArray();
-        static::assertCount(1, $logs);
-
-        static::assertSame($logs[0]['code'], 'SWAG_MIGRATION__INVALID_EMAIL_ADDRESS');
+        static::assertNotNull($convertResult->getConverted());
```
MalteJanz: We could at least check that the email field isn't there (or is it there but contains invalid data?). If it is there with invalid data then I think we can remove this entire test case as it should be covered by the validation afterwards (do we have a test for that there?).

Maybe also test that other data was successfully converted

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
```diff
@@ -171,12 +166,8 @@ public function testConvertWithoutRequiredProperties(string $property, ?string $
             $context,
             $this->migrationContext
         );
-        static::assertNull($convertResult->getConverted());
 
-        $logs = $this->loggingService->getLoggingArray();
-        static::assertCount(1, $logs);
-
-        static::assertSame($logs[0]['code'], 'SWAG_MIGRATION_EMPTY_NECESSARY_FIELD');
+        static::assertNotNull($convertResult->getConverted());
```
MalteJanz: see above, we might even remove this test case as it's not the job of the converter anymore to check for required properties but we need this to be covered by the validation layer

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
```diff
@@ -286,12 +277,7 @@ public function testConvertCustomerWithoutAddresses(): void
             $this->migrationContext
         );
 
-        static::assertNull($convertResult->getConverted());
-
-        $logs = $this->loggingService->getLoggingArray();
-        static::assertCount(1, $logs);
-
-        static::assertSame($logs[0]['code'], 'SWAG_MIGRATION_EMPTY_NECESSARY_FIELD');
+        static::assertNotNull($convertResult->getConverted());
```
MalteJanz: see above

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
```diff
@@ -309,14 +295,7 @@ public function testConvertCustomerWithoutValidAddresses(): void
...
 
-        static::assertNull($convertResult->getConverted());
-
-        $logs = $this->loggingService->getLoggingArray();
-        static::assertCount(3, $logs);
-
-        static::assertSame($logs[0]['code'], 'SWAG_MIGRATION_EMPTY_NECESSARY_FIELD');
-        static::assertSame($logs[1]['code'], 'SWAG_MIGRATION_EMPTY_NECESSARY_FIELD');
-        static::assertSame($logs[2]['code'], 'SWAG_MIGRATION_EMPTY_NECESSARY_FIELD');
+        static::assertNotNull($convertResult->getConverted());
```
MalteJanz: see above

MalteJanz (commented):
LGTM for the prototype, nice job 👍

phpstan.neon.dist
DennisGarding: Rule does not exist anymore

src/Migration/Writer/MigrationFix/MigrationFix.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $pathArray = explode(self::PATH_SEPARATOR, $this->path);
+
+        $temp = &$item;
+        foreach ($pathArray as $key) {
+            $temp = &$temp[$key];
+        }
+
+        $temp = \json_decode($this->value, true, 512, \JSON_THROW_ON_ERROR);
+        unset($temp);
```
vintagesucks: I understand this is done for performance reasons (and that this is a protoype), but I'd suggest adding a comment or extracting the dot path reference traversal logic to something with a clearer name (e.g. `setNestedValue`), since this pattern was not immediately obvious and took me a moment to parse during review.

src/Migration/Writer/MigrationFix/MigrationFixApplier.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @param array<int, string> $ids
+     *
+     * @return array<string, list<MigrationFix>>
+     */
+    private function getMappings(array $ids, string $connectionId): array
+    {
+        $sql = <<<'SQL'
+SELECT mapping.entity_uuid as entityId, fix.id, fix.value, fix.path FROM swag_migration_mapping as mapping
+INNER JOIN swag_migration_fix as fix ON fix.main_mapping_id = mapping.id
```
MalteJanz, line 55: thinking about this SQL join condition and performance: if we would store the `entity_uuid` (the UUID that the entity will have in the target SW6 system) inside the `swag_migration_fix` table instead of `main_mapping_id`, we wouldn't need a join at all.

But before we change that again, we should talk about it in our meeting tomorrow, how `swag_migration_logging` and `swag_migration_fix` relate to each other and how the latter gets inserted by the UI / bulk insert backend endpoint and how it is read again to apply the fixes.

MalteJanz (approved):

vintagesucks (approved):
LGTM

src/Migration/Writer/MigrationFix/MigrationFix.php
DennisGarding: add comment and rename temp to nestedPointer

PR #66, opened 2025-10-01
feat: convert business customers

### LINKED ISSUE #12369: Commercial customers are flagged as private customers after migrating to Shopware 6
### Shopware Version

6.7.2.0

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

If a customer is marked as commercial in Shopware 5 incl. VatID etc., he is treated as private customer after migrating from 5 to 6.

Duplicate of https://github.com/shopware/shopware/issues/3957 (Closed as not planned)

### Expected behaviour

The customer type is set to “business” after migration if one or more customer addresses has maintained company data in Shopware 5.

### How to reproduce

- Set up a Shopware 5 instance and create a commercial customer.
- Set up a Shopware 6 instance and migrate the data from the shopware 5 instance.

### Definition of Done

- [ ] An automated test covers the fix.
- [ ] Integration/E2E testing in staging is done.
- [ ] Changelog markdown file created or updated in /changelog/_unreleased directory.
- [ ] Developer documentation written/updated.
- [ ] End user and/or developer documentation written or updated with clarification.
- [ ] Regression test created to ensure bug does not reappear, and automated regression tests pass.
- [ ] Code changes resolve the reported bug without introducing new issues.

jozsefdamokos:
fixes [#12369](https://github.com/shopware/shopware/issues/12369)


src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -754,4 +767,41 @@ protected function checkEmailValidity(string $email): bool
...
+     */
+    private function getAccountType(array $originalData): string
+    {
+        $defaultBillingAddress = $this->getDefaultBillingAddress($originalData);
+
+        if ($defaultBillingAddress === null) {
+            return CustomerEntity::ACCOUNT_TYPE_PRIVATE;
+        }
+
+        if (isset($defaultBillingAddress['company']) && $defaultBillingAddress['company'] !== '') {
```
jozsefdamokos: I check only for `company` because that is the only required field if the business account is selected

vintagesucks (approved):

larskemper (approved):

src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -754,4 +767,41 @@ protected function checkEmailValidity(string $email): bool
 
         return \count($errors) === 0;
     }
+
+    /**
+     * If the customer's default billing address contains a company, the account type is business, else private.
+     */
+    private function getAccountType(array $originalData): string
+    {
+        $defaultBillingAddress = $this->getDefaultBillingAddress($originalData);
```
MalteJanz: This method is called twice if the type returned here is business right? Can you restructure the code so it is only called once.

E.g. move all the above logic (in the `convert` method) into it's own method that includes the logic of this method

src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -754,4 +756,49 @@ protected function checkEmailValidity(string $email): bool
...
+     * If the customer's default billing address contains a company, the account type is business, else private.
+     *
+     * @param array<string, mixed> $data
+     * @param array<string, mixed> $converted
+     */
+    private function setAccountType(array $data, array &$converted): void
+    {
+        $converted['accountType'] = CustomerEntity::ACCOUNT_TYPE_PRIVATE;
+
+        $defaultBillingAddress = $this->getDefaultBillingAddress($data);
```
mitelg: I would check that again, if this really should be made dependent on the billing address. while it sounds naturally, that billing address is used, afaik the address where the products are shipped to is relevant. 

at least this is what I remembered from the SW5 times. we had several issues, where we used the wrong value from the wrong address type, which needed to be fixed :grimacing:

jozsefdamokos: Hmm ok, do you have any idea who can confirm that?

mitelg: in doubt, I would contact our legal team

MalteJanz: Intuitively billing address sounds the most reasonable. But we could make the condition even wider and check the default shipping address as well.

In the end all of this is just a heuristic for detecting SW5 business customers, so they can be flagged as such in SW6 as well. Unfortunately in SW6 the decision between business and private customer is on the customer entity itself and not on the addresses like in SW5, which makes it rather difficult to find a migration approach for every scenario.

We try to cover the most common / 90% of cases here, where the customer in SW5 used a business address as their default and thus likely expects to stay a business customer in SW6 as well. If he then suddenly wants to order something private, he would need a new account for that (it's a limitation of SW6)...

I think making him "private" when he expects to be a business customer is much worse, as he likely can't do anything about it himselves and this is where the issue initially was reported to us: https://github.com/shopware/shopware/issues/12369

TLDR: we can also check both default billing and shipping addresses to find out if any of them are "business" ones and the flag the customer accordingly 🤷

PR #67, opened 2025-10-01
fix: phpstan errors

jozsefdamokos:


tests/Profile/Shopware55/Converter/PropertyGroupOptionConverterTest.php
```diff
@@ -173,6 +173,7 @@ public function testConvertWithPropertiesAndProductConfigurators(): void
 
             static::assertNotNull($converted);
             static::assertSame($convertedMainProduct['id'], $converted['id']);
+            /** @phpstan-ignore-next-line */
```
larskemper: I fixed some of these in our feature branch, you could just copy and past them. Also the other fixes, so we have less pain with rebasing the feature branch

larskemper (approved):

PR #70, opened 2025-10-07
fix: remove import from excluded test namespace

### LINKED ISSUE #12727: SW6->SW6 (SaaS) migration fails
### Shopware Version

cloud

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

SW6 -> SW6 (e.g. onPremise to SaaS) migrations fail with an exception like this:
```
Class "Shopware\Core\Test\TestDefaults" not found
```

Looks like we have multiple places where this `TestDefaults` class is used in the migration-assistant. We have to refactor these because some SW6 installations might not have tests inside of them (thus the class might be missing).

### Expected behaviour

Migration should run without exceptions caused by implementation bugs

### How to reproduce

Migration some SW6 onPremise (or maybe even another SaaS shop) into a SaaS shop

vintagesucks:
~~resolves https://github.com/shopware/shopware/issues/12727~~

~~related platform PR with a new PHPStan rule to avoid issues like this:~~
~~- https://github.com/shopware/shopware/pull/12882~~

related platform pr, which could render this one obsolete:
- https://github.com/shopware/shopware/pull/12898

### DIFF
--- a/src/Profile/Shopware6/Converter/SalesChannelConverter.php
+++ b/src/Profile/Shopware6/Converter/SalesChannelConverter.php
@@ -9,7 +9,6 @@ namespace SwagMigrationAssistant\Profile\Shopware6\Converter;
 
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
-use Shopware\Core\Test\TestDefaults;
 use SwagMigrationAssistant\Migration\Converter\ConvertStruct;
 use SwagMigrationAssistant\Migration\DataSelection\DefaultEntities;
 use SwagMigrationAssistant\Migration\MigrationContextInterface;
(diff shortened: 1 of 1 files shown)


src/Profile/Shopware6/Converter/SalesChannelConverter.php
```diff
@@ -29,7 +28,7 @@ protected function convertData(array $data): ConvertStruct
     {
         $converted = $data;
 
-        if ($converted['id'] === TestDefaults::SALES_CHANNEL) {
+        if ($converted['id'] === '98432def39fc4624b33213a56b8c944d') {
```
jozsefdamokos, line 31: We should put it in a constant in this class, so we know what the uuid stands for

vintagesucks:
This might not be needed, depending on the outcome of the discussion in https://github.com/shopware/shopware/pull/12882

ennasus4sun (approved):

src/Profile/Shopware6/Converter/SalesChannelConverter.php
vintagesucks: Do we still want this change, now that https://github.com/shopware/shopware/pull/12898 has been merged?

Most projects (probably?) use `shopware/core` as a dependency, which should resolve the issue in the next release.

mitelg: you should not change this

PR #71, opened 2025-10-07
feat(translation): Rename translation file to enable language fallbacks

### LINKED ISSUE #12815: Migrate all shopware extensions to use country-agnostic filenames
### Technical TODO

- We need to migrate all shopware extensions to use country-agnostic filenames (`de.json` instead of `de-DE.json` and so on).
- Probably it's enough to install all extensions, run `translation:lint-filenames --all` and push everything.
- Please also have a look at the min-versions of the plugins, since the language layer is in starting 6.7.3.0 (I guess)

#### List of Extensions
- [x] SaasRufus
- [x] SaasTemplate
- [x] SwagB2bPlatform
- [x] SwagCmsExtensions
- [x] SwagCommercial
- [x] SwagCustomizedProducts
- [x] SwagDynamicAccess
- [x] SwagExtensionStore
- [x] SwagMigrationAssistant
- [x] SwagMigrationMagento
- [x] SwagPayPal
- [x] SwagPublisher
- [x] SwagSocialShopping

❌ SwagLanguagePack (left out on purpose)

marcelbrode:
fixes shopware/shopware#12815


larskemper (approved):

composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "shopware/core": "~6.7"
+        "shopware/core": "~6.7.3"
```
dgrothaus-sw: The previous constraint allowed `shopware/core < 7.0.0`  whereas your new constraint only allows `shopware/Core < 6.8.0` Please check if this really was your intention or change to either `^6.7.3` or for better readability `>=6.7.3 < 7.0`

dgrothaus-sw (changes requested):

composer.json
marcelbrode: Good point! I'll change

dgrothaus-sw (approved):

composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "shopware/core": "~6.7"
+        "shopware/core": ">=6.7.3 < 7.0"
```
MalteJanz, line 13: @marcelbrode 
Is this a necessary change to happen for 6.7.3 and up? Or is it optional?

Because ideally we only raise our min supported version when we either upgrade to the next major or it is absolutely necessary for data migrations to work properly. Otherwise customers on versions 6.7.0.0 < 6.7.3.0 might still complain in dev support about issues we already fixed, causing extra work for us (also we don't maintain two branches for this case, only for major updates).

marcelbrode: We can adjust this, but the new system only works starting 6.7.3. But I can provide a work around

MalteJanz: If there is a rather easy workaround I would prefer that yes. But if not that's also not the end of the world, we just need to be more strict and say customers they need to update to get our fixes

PR #72, opened 2025-10-08
feat: delay step transition until all media is processed

### LINKED ISSUE #10822: Migration isn't finished when it says so, include media download in processing logic

jozsefdamokos:
fixes: [#10822](https://github.com/shopware/shopware/issues/10822)

- moved away from processing media in parallel messages
- all media processing is done in `MediaProcessingProcessor` now


larskemper (approved):

src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
vintagesucks, line 1: Do we need to remove those background processing hints in the UI, now that the transition is delayed until all media is processed?

Image

vintagesucks, line 1: Also, should this go into our feature branch instead of `trunk`? (not sure, just wondering)

vintagesucks (approved):

src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
jozsefdamokos: Can be either I guess

src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
```diff
@@ -101,6 +105,12 @@ public function __invoke(ProcessMediaMessage $message): void
 
             $this->loggingService->saveLogging($context);
         }
+
+        if ($run->getProgress() !== null) {
+            $this->updateProgress($message, $run->getProgress(), $context);
+        }
+
+        $this->messageBus->dispatch(new MigrationProcessMessage($context, $migrationContext->getRunUuid()));
```
MalteJanz: As discussed in our Retro, it would be great if you could add a PR description that explains how you solved the issue.

I'm having a hard time understanding how all these message queue dispatches of the `MigrationProcessMessage` and the `ProcessMediaMessage` interact now and I'm unsure if there could still be race conditions like this 🤔

MalteJanz: Also adding some documentation comments in code could help if we read it again in the future. Just for example see how this describes the flow of `MigrationProcessMessage`:
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Migration/Run/RunServiceInterface.php#L17-L27

And I think we need more comments like that, especially for not easy to reason about code.
I'm also not sure if this safety statement isn't violated by your changes:
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Migration/Run/SwagMigrationRunEntity.php#L113-L119

MalteJanz:
@jozsefdamokos thanks for the PR description 👍 . Some thoughts on these points from my side 😅 
   

> - moved away from processing media in parallel messages

Alright, no shame in doing so and it makes things a lot simpler for us to reason about which is better then race condition bugs on customer systems 👍 

> - changed MediaFileProcessorService to only dispatch a single ProcessMediaMessage instead of multiple messages. Only once this message is handled a new MigrationProcessMessage is dispatched. Then the processor calls the MediaFileProcessorService which then dispatches the next ProcessMediaMessage and so on.
> - moved updating the progress of the media entities from MediaProcessingProcessor to ProcessMediaHandler

Taking a step back, does that make sense to you? To me it sounds like we don't need `ProcessMediaMessage` and `ProcessMediaHandler` at all anymore and just merge the necessary logic or refactoring it into `MediaProcessingProcessor` (which is called during the handling of `MigrationProcessMessage`) or the service that it is calling `MediaFileProcessorService` (which then shouldn't dispatch messages anymore). That way we don't jump between different message types and our code is much easier to reason about I think 🤔 

> - changed each ProcessMediaMessage to hold 20 media files to process instead of the previous 5.

Also sounds reasonable to me. Did you somehow manage to measure how long the processing of these messages take with that "batch size", ideally downloading from a remote server like a Shopware build (ShopDev) instance? We should just make sure we stay well below the 30s request timeout limit (in case someone migrates with the admin worker instead of proper cli worker).

DennisGarding (approved):

jozsefdamokos:
@MalteJanz Yes, I thought about merging them, I just chose the quicker way for now to validate that this approach works. I agree, we should merge the two now if the approach of moving away from parallel processing is fine.

src/Migration/Logging/Log/CannotGetFileRunLog.php
```diff
@@ -18,7 +17,7 @@ public function __construct(
         string $entity,
         string $sourceId,
         private readonly string $uri,
-        private readonly ?RequestException $requestException = null,
```
jozsefdamokos, line 21: This can also be a `ConnectException`

DennisGarding (approved):

larskemper (approved):

tests/acceptance/tests/MigrationByUiFlow.spec.ts
```diff
@@ -102,17 +102,6 @@ test('As a shop owner I want to migrate my data from my old SW5 shop to SW6 via
         await page.getByRole('button', { name: 'Back to overview' }).click();
     });
 
-    // ToDo MIG-985: Remove this if the underlying issue is fixed
-    await test.step('Wait for media download to finish', async () => {
-        await expect.poll(async () => {
-            return await MediaProcessObserver.isMediaProcessing();
```
MalteJanz, line 108: I guess this was the only place where `MediaProcessObserver` was used? So you can remove that file as well:
https://github.com/shopware/SwagMigrationAssistant/blob/17155df474c74cabd3a70aa8f5185ca4e312580f/tests/acceptance/fixtures/MediaProcessObserver.ts

MalteJanz (approved):
LGTM 👍 , only found one more file that can be removed now I think 🙂

PR #76, opened 2025-10-23
12367/add entity classes

DennisGarding:
Close: https://github.com/shopware/shopware/issues/12367

This PR adds the required entity classes.

Adds a normally not required test for the entity. In this case its required, because in the get and set value methods is logic


src/Core/Migration/Migration1757598733AddMigrationFixesTable.php
```diff
@@ -17,13 +17,13 @@
 #[Package('fundamentals@after-sales')]
 class Migration1757598733AddMigrationFixesTable extends MigrationStep
 {
-    public const MIGRATION_FIXES_TABLE = 'swag_migration_fixes';
+    public const MIGRATION_FIXES_TABLE = 'swag_migration_fix';
```
DennisGarding, line 20: I think we should rename the table to give it a consistent, singular name.

src/Core/Migration/Migration1757598733AddMigrationFixesTable.php
```diff
@@ -17,13 +17,13 @@
...
 
     public const FIELDS = [
         'id' => 'BINARY(16) NOT NULL',
         'connection_id' => 'BINARY(16) NOT NULL',
         'main_mapping_id' => 'BINARY(16) NOT NULL',
-        'value' => 'JSON NOT NULL',
+        'value' => 'LONGTEXT NOT NULL',
```
DennisGarding: As i mentioned, the JsonField of shopware does only accept an array. We have different value types

MalteJanz: Considering that the getter and setter of the entity always expects to read / write JSON, I think it would still be better to enforce that on a DB level and keep the `json` type here. Otherwise the admin api might write some non json value into the DB and then PHP tries to read and fail suddenly. It would be better to fail early (during write) than late (during read).

Even better would be going with a custom DAL field type with it's own serializer that validates the data

tests/integration/Migration/MigrationFix/SwagMigrationEntityTest.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public static function valueData(): array
+    {
+        return [
+            'integer' => ['value' => 12],
+            'float' => ['value' => 42.12],
+            'string' => ['value' => 'fooBar'],
+            'array' => ['value' => ['key' => 'value']],
+        ];
+    }
```
MalteJanz, line 88: Isn't the value that is returned here and passed as an argument to the above method always an array like `['value' => 12]` instead of just a number `12`?

MalteJanz: Also thinking further about it: we could also go with a `JsonField` DAL field and make a convention to always set json objects like
```json
{
    "value": ...
}
```
and accessing it always under `value`, then it also could by any type, including an array or object

src/Core/Migration/Migration1757598733AddMigrationFixesTable.php
MalteJanz: see also my other comment here: https://github.com/shopware/SwagMigrationAssistant/pull/76#discussion_r2455390732

tests/integration/Migration/MigrationFix/SwagMigrationEntityTest.php
DennisGarding: No, value is a named argument. the mixed $value variable has only the value 12, 42.4, 'foobar' and ['key' => 'value']

DennisGarding: I think this makes it much more complicated. Also, we use plain SQL to get the fix.

MalteJanz: What are your arguments for thinking it would be more complicated always going with a json object containing a single key `value`?

I think the current approach also has some downsides and relies on the setters, which in turn also makes stuff complicated

DennisGarding: You always have to get the value out of the object. Thats unnecessary and complicate. If you dont know that, you have to investigate further.

I like the idea with the DAL Field extension. This should do the work for us

src/Migration/MigrationFix/SwagMigrationFixDefinition.php
```diff
@@ -0,0 +1,54 @@
...
+        return SwagMigrationFixEntity::class;
+    }
+
+    protected function defineFields(): FieldCollection
+    {
+        return new FieldCollection([
+            (new IdField('id', 'id'))->addFlags(new PrimaryKey(), new Required()),
+            (new IdField('connection_id', 'connectionId'))->addFlags(new Required()),
+            (new IdField('main_mapping_id', 'mainMappingId'))->addFlags(new Required()),
+            (new MigrationJsonField('value', 'value'))->addFlags(new Required()),
```
MalteJanz: I'm not sure if the `value` field really should be required, because a valid fix could also say set the field with this path to `null`, e.g. if it is an optional field, contained a broken value before and you prefer to set it to nothing instead of some unnecessary dummy data.

But maybe it already works because `null` is also a valid json value I think, see https://www.w3schools.com/js/js_json_datatypes.asp

src/Core/Field/MigrationJsonField.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Shopware\Core\Framework\DataAbstractionLayer\Field\Field;
+use Shopware\Core\Framework\DataAbstractionLayer\Field\StorageAware;
+use Shopware\Core\Framework\Log\Package;
+
+/**
+ * @internal
+ */
+#[Package('after-sales')]
+class MigrationJsonField extends Field implements StorageAware
```
MalteJanz: Nice one, I like this approach 👍 .

I only would reconsider the name of the field, because it has nothing to do with "migration" and is already in the migration plugin namespace. Maybe rename it to something like `AnyJsonField` and add a doc comment above it that it allows any json value (e.g. strings, numbers, objects, arrays) instead of just objects (what the normal JsonField allows)

src/Migration/MigrationFix/SwagMigrationFixDefinition.php
DennisGarding: Then i had to remove the validation. But its ok i think. We just want to save anything and load it again!

MalteJanz (approved):

PR #79, opened 2025-10-30
refactor: remove validation from converter

DennisGarding:
Remove all checks, which results in an empty Convert Result.  These should be catches by the later entity validation


MalteJanz:
Please add a short PR description 🙂 . E.g. questions I would like to see answered in it:
- What type of checks are being removed, e.g. checks for entity fields that are required in SW6 and would produce another log entry otherwise in the later entity validation
- What other type of checks are removed and what's the reasoning behind it. E.g. does the user still get notified about errors in optional fields? Or are they just silently set to `null`?

src/DependencyInjection/shopware.xml
```diff
@@ -212,7 +212,6 @@
```
larskemper, line 212: Same here please use conventional commit style for the pr title:
https://media.cheatography.com/storage/thumb/albelop_conventional-commits.750.jpg

larskemper (changes requested):

src/Profile/Shopware/Converter/CustomerConverter.php
```diff
@@ -404,13 +398,13 @@ protected function applyAddresses(array &$originalData, array &$converted, strin
         $converted['addresses'] = $addresses;
 
         // No valid default billing and shipping address was converted, so use the first valid one as default
-        $this->checkUnsetDefaultShippingAndDefaultBillingAddress($originalData, $converted, $customerUuid, $addresses);
+        $this->checkUnsetDefaultShippingAndDefaultBillingAddress($originalData, $converted, $addresses);
```
ennasus4sun, line 401: looks like we could remove $customerUuid as a param, right?

DennisGarding: It is already done. I think you reviewed just an interim result

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
```diff
@@ -138,6 +139,40 @@ public function testConvert(): void
...
+        static::assertNotNull($convertResult->getConverted());
+    }
+
+    /**
+     * @return list<array{0: string, 1: string|null}>
+     */
+    public static function requiredProperties(): array
+    {
+        return [
+            ['email', null],
```
ennasus4sun: I know that is working, but for me using yield/Generators + using array-keys like
```
 ['property' => 'email', 
'value' => null]
```
  is really more readable
some more advantages for the use of Gnerators can be:

- named test cases > finding the failing test
- you can run them individually > makes debugging easier

ennasus4sun (commented):
nice 💪🏻

tests/Profile/Shopware55/Converter/CustomerConverterTest.php
DennisGarding: I think you still review an old interim stand... 

never mind. i am on the wrong branch :sweat_smile:

src/Profile/Shopware/Converter/CustomerConverter.php
ennasus4sun: ok, I still see it ... 😕 
just to be sure we talk about the same:
`$customerUuid` passed to this function (`applyAddresses`) is now not used anymore
image

ennasus4sun (approved):

larskemper (approved):

src/Migration/Writer/SeoUrlWriter.php
```diff
@@ -50,6 +50,11 @@ public function supports(): string
      */
     public function writeData(array $data, Context $context): array
     {
+        // only unmodified should be written
```
MalteJanz: What's the reasoning behind this?

src/Profile/Shopware/Converter/CategoryConverter.php
```diff
@@ -84,19 +82,11 @@ public function convert(
...
-                    ->withFieldSourcePath('_locale')
-                    ->withSourceData($data)
-                    ->build(EmptyNecessaryFieldRunLog::class)
-            );
+        $converted = [];
 
-            return new ConvertStruct(null, $data);
+        if (isset($data['_locale'])) {
+            $this->locale = $data['_locale'];
         }
```
MalteJanz, line 91: I would add an `else` branch here and set `$this->locale` to `null`. Otherwise the value could stay for the next call to convert, which might cause a bug that is really hard to reason about.

Previously this wasn't an issue because of the early return and always setting it to the value of `$data['_locale']`, which after the early return always exists.

src/Profile/Shopware/Converter/CategoryConverter.php
```diff
@@ -245,7 +235,7 @@ protected function setGivenCategoryTranslation(array &$data, array &$converted):
 
         $locale = $language->getLocale();
 
-        if ($locale === null || $locale->getCode() === $data['_locale']) {
+        if (!isset($data['_locale']) || $locale === null || $locale->getCode() === $data['_locale']) {
```
MalteJanz: It would also make this easier, just checking for `$locale === null || $locale->getCode() === $this->locale` would be enough. But I'm a bit confused what this condition does in the first place, is it correct to early return if the language has no locale or the locale code matches the one of data (the latter sounds confusing)? 🤔

src/Profile/Shopware/Converter/CategoryConverter.php
```diff
@@ -346,7 +336,7 @@ protected function addMediaTranslation(array &$media, array $data): void
         }
 
         $locale = $language->getLocale();
-        if ($locale === null || $locale->getCode() === $this->locale) {
+        if ($locale === null || $this->locale === null || $locale->getCode() === $this->locale) {
```
MalteJanz: see above, also the added check seems unnecessary, the strict equal should be enough

src/Profile/Shopware/Converter/CategoryConverter.php
```diff
@@ -364,10 +354,12 @@ protected function addMediaTranslation(array &$media, array $data): void
         $localeTranslation['id'] = $mapping['entityId'];
         $this->mappingIds[] = $mapping['id'];
 
-        $languageUuid = $this->languageLookup->get($this->locale, $this->context);
-        if ($languageUuid !== null) {
-            $localeTranslation['languageId'] = $languageUuid;
-            $media['translations'][$languageUuid] = $localeTranslation;
+        if ($this->locale !== null) {
+            $languageUuid = $this->languageLookup->get($this->locale, $this->context);
```
MalteJanz, line 359: See above `$this-locale` might contain the wrong value now if not adjusted

MalteJanz (changes requested):
Still have to look further (didn't had the time yet for a full review) 🙈 .

And guys, please review and test this very carefully. Because we can easily introduce new bugs here, especially as our acceptance tests aren't running right now in the feature branch https://github.com/shopware/SwagMigrationAssistant/blob/6c36b6d62ce19debe052afa423f09506577a9976/tests/acceptance/tests/MigrationByUiFlow.spec.ts#L17-L19

src/Migration/Writer/SeoUrlWriter.php
DennisGarding: This seems to be a logic issue. This should be the other way around. 

https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Profile/Shopware6/Converter/SeoUrlConverter.php#L38C9-L40C10

src/Profile/Shopware/Converter/CategoryConverter.php
DennisGarding: you right. This test should be enough.

jozsefdamokos (approved):

src/Migration/Writer/SeoUrlWriter.php
DennisGarding: Change comment to modified should only be written

DennisGarding: Look into SeoUrlConverter. This is only SW6 behavior. Maybe early return in converter

src/Migration/Converter/ConverterInterface.php
```diff
@@ -40,7 +40,7 @@ public function getMediaUuids(array $converted): ?array;
      *
      * @param array<string, mixed> $data
      */
-    public function convert(array $data, Context $context, MigrationContextInterface $migrationContext): ConvertStruct;
+    public function convert(array $data, Context $context, MigrationContextInterface $migrationContext): ?ConvertStruct;
```
MalteJanz, line 45: Nitpick: would be nice if the use case for `null` return is explained in the doc comment above

src/Profile/Shopware/Converter/CurrencyConverter.php
```diff
@@ -49,23 +49,6 @@ public function convert(array $data, Context $context, MigrationContextInterface
...
-                    DefaultEntities::CURRENCY,
-                    $data['currency'],
-                    $this->checksum,
-                    null,
-                    $currencyUuid
-                );
-            }
-
-            return new ConvertStruct(null, $data);
-        }
```
MalteJanz, line 67: I think this code was actually important, when the currency already exists in SW6, there is no need to create it again. So creating the mapping based on the lookup should still happen and ideally also the early return that this entity was skipped (because it already exists)

src/Profile/Shopware/Converter/LanguageConverter.php
```diff
@@ -51,15 +49,13 @@ public function convert(array $data, Context $context, MigrationContextInterface
...
-            );
-
-            return new ConvertStruct(null, $data);
+            $this->loggingService->addLogEntry(new EntityAlreadyExistsRunLog(
+                $migrationContext->getRunUuid(),
+                DefaultEntities::LANGUAGE,
+                $data['id']
+            ));
+
+            return null;
```
MalteJanz: Is this producing two logs now? I think when `null` is returned you also produce a log entry on the caller side of `convert` right?

src/Profile/Shopware/Converter/MainVariantRelationConverter.php
```diff
@@ -34,10 +32,6 @@ public function convert(array $data, Context $context, MigrationContextInterface
         $connection = $migrationContext->getConnection();
         $this->connectionId = $connection->getId();
 
-        if (!isset($data['id'], $data['ordernumber'])) {
-            return new ConvertStruct(null, $data);
```
MalteJanz, line 38: If these aren't set, the calls below for the main mapping lookup will fail (because access of undefined array key). So In the unlikely and very special case that we don't receive an `id` (+ in this case `orderNumber`) from the source system, we can't really convert this entity, because we don't have a mapping and SW6 uuid to write it later to.

TLDR: a early return for these special cases is fine, but should be logged with an error.

src/Profile/Shopware6/Converter/MediaFolderConverter.php
```diff
@@ -49,48 +49,46 @@ protected function convertData(array $data): ConvertStruct
...
-        if ($converted['id'] === str_repeat('0', 32)) {
-            // edge case for hidden download products media folder
-            // this doesn't need to be migrated, just map the files into it
-            // otherwise the code below would lead to a self-referential parentId
-            return new ConvertStruct(null, $data, $this->mainMapping['id'] ?? null);
-        }
+        // edge case for hidden download products media folder
+        // this doesn't need to be migrated, just map the files into it
+        // otherwise the code below would lead to a self-referential parentId
+        if ($converted['id'] !== str_repeat('0', 32)) {
```
MalteJanz: This edge case should still not be migrated, thus needs a `return null` early return

tests/_fixtures/Shopware6/CustomerWishlist/02-CustomerNotFound/output.php
```diff
@@ -5,4 +5,35 @@
  * file that was distributed with this source code.
  */
 
-return null;
+return [
+    'customerId' => '872e88b6d30048e388b018f7a0be6a1e',
```
MalteJanz: If the customer couldn't be found, why does this entity have a `customerId` then 🤔

tests/_fixtures/Shopware6/CustomerWishlist/03-SalesChannelNotFound/output.php
```diff
@@ -5,4 +5,35 @@
  * file that was distributed with this source code.
  */
 
-return null;
+return [
+    'customerId' => '872e88b6d30048e388b018f7a0be6a1e',
+    'salesChannelId' => 'a8084cf2687b44db91f3277b57cc3d7a',
```
MalteJanz: see above, where is this `salesChannelId` coming from

tests/_fixtures/Shopware6/MediaFolder/04-DigitalProductsMediaFolder/output.php
```diff
@@ -5,4 +5,20 @@
  * file that was distributed with this source code.
  */
 
-return null;
+return [
+    'id' => '00000000000000000000000000000000',
```
MalteJanz: this should still return `null`, see my comment above and the code comment in the converter. This special folder shouldn't be migrated, because it already exists in every SW6 install

tests/Profile/Shopware55/Converter/CategoryConverterTest.php
```diff
@@ -132,9 +132,9 @@ public function testConvertWithParent(): void
             new DummyMappingService(),
             $this->loggingService,
             new DummyMediaFileService(),
-            $this->getContainer()->get(LowestRootCategoryLookup::class),
-            $this->getContainer()->get(DefaultCmsPageLookup::class),
-            //            $this->getContainer()->get(LanguageLookup::class),
+            static::getContainer()->get(LowestRootCategoryLookup::class),
+            static::getContainer()->get(DefaultCmsPageLookup::class),
+            //            static::getContainer()->get(LanguageLookup::class),
```
MalteJanz: Nitpick: can we remove this comment line?

MalteJanz (approved):
After my open comments are solved: LGTM, nice job, this was quite a task 💪

PR #80, opened 2025-10-31
Chore: 13247/change table swag migration logging

DennisGarding:
Closes: [#80 ](https://github.com/shopware/shopware/issues/13247)

- add relation to "entity_id"
- add index to "entity_id"
- remove field "used_mapping"


src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function getCreationTimestamp(): int
+    {
+        return 1761903189;
+    }
+
+    public function update(Connection $connection): void
+    {
+        if ($this->columnExists($connection, 'swag_migration_logging', 'used_mapping')) {
+            $connection->executeStatement('ALTER TABLE `swag_migration_logging` DROP COLUMN `used_mapping`;');
```
MalteJanz: Sometimes it helps to look at the parent class what utilities are already available before implementing your own 😉

https://github.com/shopware/shopware/blob/9f97e99531f360d220a3a667e141a9cd0a00bcac/src/Core/Framework/Migration/MigrationStep.php#L100-L103

src/Core/Migration/TableHelperTrait.php
```diff
@@ -0,0 +1,25 @@
...
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Migration\ColumnExistsTrait;
+
+#[Package('after-sales')]
+trait TableHelperTrait
+{
+    use ColumnExistsTrait;
+
+    protected function indexExists(Connection $connection, string $table, string $indexName): bool
```
MalteJanz: In what way is this different from 
https://github.com/shopware/shopware/blob/9f97e99531f360d220a3a667e141a9cd0a00bcac/src/Core/Framework/Migration/MigrationStep.php#L84 ?

src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        if (!$this->columnExists($connection, 'swag_migration_logging', 'entity_id')) {
+            $connection->executeStatement('ALTER TABLE `swag_migration_logging` ADD COLUMN `entity_id` BINARY(16) NULL;');
```
MalteJanz: Take a look at:
https://github.com/shopware/shopware/blob/82692c9aa7f8e9349f67fe267b566ea04e55570a/src/Core/Framework/Migration/AddColumnTrait.php#L11-L21

which is already in scope because that trait is used by `MigrationStep`

src/Migration/Logging/SwagMigrationLoggingEntity.php
```diff
@@ -49,13 +49,10 @@ class SwagMigrationLoggingEntity extends Entity
...
     protected ?array $convertedData = null;
 
-    /**
-     * @var array<string, mixed>|null
-     */
-    protected ?array $usedMapping = null;
-
     protected ?string $exceptionMessage = null;
 
+    protected string $entityId;
```
MalteJanz: the field is nullable right?
```suggestion
    protected ?string $entityId;
```

src/Migration/Logging/SwagMigrationLoggingEntity.php
```diff
@@ -244,4 +225,14 @@ public function setExceptionTrace(array $exceptionTrace): void
     {
         $this->exceptionTrace = $exceptionTrace;
     }
+
+    public function getEntityId(): string
```
MalteJanz:
```suggestion
    public function getEntityId(): ?string
```

src/Migration/Logging/SwagMigrationLoggingEntity.php
```diff
@@ -244,4 +225,14 @@ public function setExceptionTrace(array $exceptionTrace): void
...
+    {
+        return $this->entityId;
+    }
+
+    public function setEntityId(string $entityId): void
```
MalteJanz:
```suggestion
    public function setEntityId(?string $entityId): void
```

src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
DennisGarding: Could be helpful, if these helper functions are extracted and prominent placed. :man_shrugging:

src/Core/Migration/TableHelperTrait.php
DennisGarding: No diff in its function.. Good to know

src/Migration/Logging/SwagMigrationLoggingEntity.php
DennisGarding: yes.. in the first step i make it required, but it turned out its not possible

src/Migration/Run/RunService.php
```diff
@@ -248,26 +248,28 @@ public function assignThemeToSalesChannel(string $runUuid, Context $context): vo
...
             try {
-                $this->themeService->assignTheme($defaultTheme, $salesChannel, $context);
+                $this->themeService->assignTheme($defaultThemeId, $salesChannelId, $context);
             } catch (\Throwable $exception) {
                 $this->loggingService->addLogEntry(
                     (new SwagMigrationLogBuilder(
                         $runUuid,
                         $connection->getProfileName(),
                         $connection->getGatewayName(),
+                        $salesChannelId,
```
MalteJanz: Why pass in the `entityId` twice? I would remove this constructor argument and just use the `withEntityId` method to make it more explicit.

src/Migration/Run/RunService.php
```diff
@@ -248,26 +248,28 @@ public function assignThemeToSalesChannel(string $runUuid, Context $context): vo
...
                     ))
                         ->withExceptionMessage($exception->getMessage())
                         ->withExceptionTrace($exception->getTrace())
                         ->withEntityName(SalesChannelDefinition::ENTITY_NAME)
+                        ->withEntityId($salesChannelId)
                         ->build(ThemeCompilingErrorRunLog::class)
```
MalteJanz: This is where things get complicated. The `ThemeCompilingErrorRunLog` has `isUserFixable` `false`, so it isn't fixable by the user anyways and we just have to think about what information is the most useful for this type of error.

And considering the name of the error I don't think giving you the reference of the saleschannel helps much, instead I would prefer getting details of the theme I think, so something like
```
->withEntityName(ThemeDefinition::ENTITY_NAME)
->withEntityId($defaultThemeId)
```

Or do you think the salechannel is the part that causes issues here? Then I would at least rename the log to something like `ThemeSalesChannelAssignmentErrorRunLog` 🤷

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -70,12 +71,12 @@ public function validate(MigrationContextInterface $migrationContext, Context $s
             $this->validateFields($context);
             $this->validateAssociations($context);
         } catch (\Throwable $e) {
-            $this->loggingService->addLogEntry(
+            $context->getValidationResult()->addLog(
                 SwagMigrationLogBuilder::fromMigrationContext($context->getMigrationContext())
                     ->withEntityName($context->getEntityDefinition()->getEntityName())
```
MalteJanz: I would still try to add the entityId here if available:
```php
->withEntityId($context->getConvertedData()['id'] ?? null)
```

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -135,6 +138,10 @@ private function validateFields(SwagMigrationValidationContext $context): void
             throw MigrationException::unexpectedNullValue('id');
         }
 
+        if (!Uuid::isValid($context->getConvertedData()['id'])) {
+            throw MigrationException::invalidId($context->getConvertedData()['id'], $context->getEntityDefinition()->getEntityName());
```
MalteJanz: Good catch, this is an important check, for when we return data from a converter that doesn't contain a primary `id`, which is an implementation error by us.

But thinking about how would you troubleshoot that error given the information? Because `id` is likely `null` and you only have the `EntityName`. I think in this case you really need a way to figure out what data was passed to the converter (e.g. the source data). Would be nice if you could pass that somehow to this validation service as well and include it with the logs 🤔

src/Migration/Logging/Log/Builder/SwagMigrationLogBuilder.php
```diff
@@ -72,6 +71,13 @@ public function withFieldSourcePath(string $fieldSourcePath): self
         return $this;
     }
 
+    public function withEntityId(?string $entityId): self
```
larskemper, line 75:
```suggestion
    public function withEntityId(string $entityId): self
```

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -70,12 +71,12 @@ public function validate(MigrationContextInterface $migrationContext, Context $s
             $this->validateFields($context);
             $this->validateAssociations($context);
         } catch (\Throwable $e) {
-            $this->loggingService->addLogEntry(
+            $context->getValidationResult()->addLog(
                 SwagMigrationLogBuilder::fromMigrationContext($context->getMigrationContext())
                     ->withEntityName($context->getEntityDefinition()->getEntityName())
                     ->withExceptionMessage($e->getMessage())
                     ->withExceptionTrace($e->getTrace())
```
MalteJanz: Also adding `convertedData` if it's available could be helpful to troubleshoot the exception:
```php
->withConvertedData($context->getConvertedData())
```

larskemper (approved):

MalteJanz (changes requested):
Are these already all places where logs are created? I guess any logs that are created in converters are being covered by your other PR then? https://github.com/shopware/SwagMigrationAssistant/pull/79

src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
MalteJanz: It's already super easy to find them. Just use your autocomplete on `$this->` or browse the parent classes a bit. If you place them in some Trait / Service class you are more likely to overlook them.

So I'm curious how your idea looks like to fulfill 

> helper functions are extracted and prominent placed

But you are not the first one that ran into this. I guess @larskemper also ran into this recently if I remember correctly ✌️ 🙈. So I'm curious on how the developer experience can be improved here 🙂

src/Migration/Validation/SwagMigrationValidationService.php
DennisGarding: This error is catches and added to the log 

custom/plugins/SwagMigrationAssistant/src/Migration/Service/MigrationDataConverter.php:117

src/Migration/Run/RunService.php
DennisGarding: I think your right.

src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
DennisGarding: i searched for columnExists and find the trait "ColumnExistsTrait".. In this directory i searched for IndexExists and did not find that.  :man_shrugging: 

As the saying goes: If you know you know :smile:

src/Migration/Validation/SwagMigrationValidationService.php
MalteJanz: I guess this particular throw would already be catched here:
https://github.com/shopware/SwagMigrationAssistant/blob/1f6964c32a3d011dd1025451bc8e9bb92e3b516f/src/Migration/Validation/SwagMigrationValidationService.php#L73-L81

And then I would add the additional information there

DennisGarding: Thats what i mean. Followed the wrong path.

DennisGarding: Converted data is already added

src/Core/Migration/Migration1761903189AdjustSwagMigrationLoggingTable.php
MalteJanz: Where are you searching?

image

image

src/Migration/Validation/SwagMigrationValidationService.php
MalteJanz: If you don't have an `id` you likely don't have converted data (or it's empty), so I would still look into if it is possible to add the source data as well for these `ValidationExceptionLog` cases

src/Core/Migration/TableHelperTrait.php
MalteJanz: Can this trait be removed then?

src/Migration/Logging/SwagMigrationLoggingEntity.php
MalteJanz: This isn't resolved yet. Any reason for that?

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -194,6 +195,7 @@ private function handleWriteException(
                     ->withExceptionTrace($exception->getTrace())
                     ->withEntityName($entityName)
                     ->withConvertedData($entity)
+                    ->withEntityId($entity['id'] ?? null)
```
MalteJanz, line 206: Because you changed the builder method to only accept a `string` this shouldn't work anymore. But to be honest I prefer this way of setting it instead of wrapping the call to `->withEntityId` in an `if` branch. I leave it up to you

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -245,13 +247,18 @@ private function writePerEntity(
                 $currentWriter = $this->writerRegistry->getWriter($entityName);
                 $currentWriter->writeData([$entity], $this->writeContext);
             } catch (\Throwable $exception) {
+                $logBuilder = SwagMigrationLogBuilder::fromMigrationContext($migrationContext)
+                    ->withExceptionMessage($exception->getMessage())
+                    ->withExceptionTrace($exception->getTrace())
+                    ->withEntityName($entityName)
+                    ->withConvertedData([$entity]);
```
MalteJanz: As stated in another comment, would be great to add the source data here as well to better troubleshoot the exception

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -110,6 +117,7 @@ private function validateEntityStructure(SwagMigrationValidationContext $context
                     ->withEntityName($context->getEntityDefinition()->getEntityName())
                     ->withFieldName($missingField)
                     ->withConvertedData($context->getConvertedData())
+                    ->withEntityId($context->getConvertedData()['id'] ?? null)
```
MalteJanz: see above

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -122,6 +130,7 @@ private function validateEntityStructure(SwagMigrationValidationContext $context
                     ->withEntityName($context->getEntityDefinition()->getEntityName())
                     ->withFieldName($unexpectedField)
                     ->withConvertedData($context->getConvertedData())
+                    ->withEntityId($context->getConvertedData()['id'] ?? null)
```
MalteJanz: see above

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -172,6 +185,7 @@ private function validateFields(SwagMigrationValidationContext $context): void
                         ->withConvertedData([$fieldName => $value])
                         ->withExceptionMessage($e->getMessage())
                         ->withExceptionTrace($e->getTrace())
+                        ->withEntityId($context->getConvertedData()['id'] ?? null)
```
MalteJanz: see above

src/Migration/Validation/SwagMigrationValidationService.php
```diff
@@ -223,6 +237,7 @@ private function validateAssociations(SwagMigrationValidationContext $context):
                         ->withEntityName($context->getEntityDefinition()->getEntityName())
                         ->withFieldName($fkFieldName)
                         ->withConvertedData([$fkFieldName => $fkValue])
+                        ->withEntityId($context->getConvertedData()['id'] ?? null)
```
MalteJanz: see above

src/Core/Migration/TableHelperTrait.php
MalteJanz: Nevermind, you still use it in the test. But maybe move this trait to the tests folder then? Because otherwise it might be accidentally used in normal migrations, even though that isn't necessary

DennisGarding: No.. I added needed functionality , moved to the test directory and used it for the tests.

src/Migration/Logging/SwagMigrationLoggingEntity.php
DennisGarding: Yes... Phpstan says NO! 

`Parameters should have "string" types as the only types passed to this method  
         🪪  typePerfect.narrowPublicClassMethodParamType`

DennisGarding: I disabled this PhpStan check! Telgi said that this setting is also off in the core

src/Core/Migration/Migration1754897550AddFieldsToMigrationLogs.php
```diff
@@ -20,6 +21,8 @@
 #[Package('fundamentals@after-sales')]
 class Migration1754897550AddFieldsToMigrationLogs extends MigrationStep
 {
+    use ColumnExistsTrait;
```
MalteJanz: This shouldn't be necessary as it's already part of `MigrationStep` (which uses `AddColumnTrait`, which itself uses `ColumnExistsTrait`)

src/Core/Migration/Migration1754897550AddFieldsToMigrationLogs.php
```diff
@@ -69,31 +82,20 @@ public function update(Connection $connection): void
...
+            if ($this->columnExists($connection, 'swag_migration_logging', $column)) {
                 $connection->executeStatement(
                     \sprintf(
                         'ALTER TABLE `%s` DROP COLUMN `%s`;',
                         self::MIGRATION_LOGGING_TABLE,
-                        $column->getName()
+                        $column
                     )
                 );
             }
```
MalteJanz: While you are touching this, would be great if this would just use the existing method:
https://github.com/shopware/shopware/blob/9f97e99531f360d220a3a667e141a9cd0a00bcac/src/Core/Framework/Migration/MigrationStep.php#L100-L103

MalteJanz (approved):

src/Core/Migration/Migration1754897550AddFieldsToMigrationLogs.php
larskemper, line 1: same here, please use conventional commit style pr title

PR #85, opened 2025-11-10
refactor: use xxh hash

larskemper:
resolves [#13391](https://github.com/shopware/shopware/issues/13391)

use `xxh128` where possible for better performance & use `Hasher` util for all hashing uses.


src/Profile/Shopware/Converter/ProductConverter.php
```diff
@@ -672,14 +673,14 @@ private function applyOptions(array &$converted, array &$data): void
             $optionMapping = $this->mappingService->getOrCreateMapping(
                 $this->connectionId,
                 DefaultEntities::PROPERTY_GROUP_OPTION,
-                \hash('md5', \mb_strtolower($option['name'] . '_' . $option['group']['name'])),
+                Hasher::hash(\mb_strtolower($option['name'] . '_' . $option['group']['name'])),
```
MalteJanz: For all the calls to `mappingService`, we still have to use `md5` and guarantee that the hash is the same as before, given the same inputs.

Because otherwise we might break shops that previously migrated (have mappings with md5 hashes) and migrate again where these mappings couldn't be found and new ones (xxh hash) are generated, which leads to duplicate data in the best case and exceptions in the worst case (e.g. because of unique constraints)

src/Profile/Shopware/Converter/PropertyGroupOptionConverter.php
```diff
@@ -49,7 +50,7 @@ public function getSourceIdentifier(array $data): string
             $group = $data['group']['name'];
         }
 
-        return \hash('md5', \mb_strtolower($data['name'] . '_' . $group . '_' . $data['type']));
+        return Hasher::hash(\mb_strtolower($data['name'] . '_' . $group . '_' . $data['type']));
```
MalteJanz: see above, this has to match the mapping lookup and needs to stay `md5`

src/Profile/Shopware/Converter/ShippingMethodConverter.php
```diff
@@ -748,7 +749,7 @@ private function setCustomAvailabilityRule(array $data, array &$converted): void
         $ruleData = $this->getRelevantDataForAvailabilityRule($data);
 
         $jsonRuleData = \json_encode($ruleData, \JSON_THROW_ON_ERROR);
-        $hash = \md5($jsonRuleData);
+        $hash = Hasher::hash($jsonRuleData);
```
MalteJanz: see above, used for mapping here. Basically all places that are used for mappings should stay `md5`. We can only change the checksum hash freely

MalteJanz (dismissed):

src/Profile/Shopware/Converter/ProductConverter.php
jozsefdamokos: But if we do it for the feature branch, which will not work with existing migrations anyway we can keep it right? Would be nice to use the same hashing across the plugin.

jozsefdamokos (approved):

src/Profile/Shopware/Converter/ProductConverter.php
larskemper: we will also release a major version, right? So we can decide on doing that

MalteJanz: Even with our feature branch / a major version you don't want to break like this. E.g. there will be customers who have the migration-assistant installed, migrated today, update to our upcoming major tomorrow, and migrate again. Sure the migration user experience / flow will change with the fixes but the migration itself still works the same, the customer wouldn't expect that he suddenly has lot's of data duplicated in his shop. Also I can't imagine we want to have that kind of dev support tickets, troubleshooting that. The only fix for them would then be to delete their shop and start fresh 🤯 .

So in the end our `swag_migration_mapping` data is really important for our customers, to continue migrating without deleting their whole shop or getting lots of stuff duplicated. Means it's really difficult for us to make changes on how the mapping works and I guess we already fucked up in the past on smaller occasions.

The only clean way of changing our mapping I can think of would be to release a completely new Migration-Assistant2 plugin and deprecate the old one, but that's not something we should consider for just replacing some hashing algorithm. Or at least make it a new migration profile that is the new default... 🤷

larskemper: ok, got it 😄. Will revert

src/Migration/Media/Processor/HttpDownloadServiceBase.php
```diff
@@ -318,13 +319,13 @@ private function persistFileToMedia(string $filePath, string $uuid, string $name
             return;
         }
 
-        $fileHash = \hash_file('md5', $filePath);
+        $fileHash = Hasher::hash(\file_get_contents($filePath));
```
ennasus4sun: I'm always a little unsure about what to use here. But Telgi usually recommends using the Symfony Filesystem Component. In any case, this makes sense in order to be able to test code better. Maybe it makes sense to use it here to instead the native PHP function

MalteJanz: There is an even easier way: just use `Hasher::hashFile(...)`. Which in turn also uses `\hash_file` internally. Regarding making it easier to test: That's already the case when we don't use any PHP functions like `\hash_file` or `\file_get_contents` directly, in this case you could just mock `Hasher::hashFile` if you don't want it to read a real file for a test.

DennisGarding (approved):

ennasus4sun (approved):

PR #91, opened 2025-11-18
fix: dont migrate core.app.shopIdV2

jozsefdamokos:
- added core.app.shopIdV2 to the blocked list


src/DataProvider/Provider/Data/SystemConfigProvider.php
```diff
@@ -26,6 +26,7 @@ class SystemConfigProvider extends AbstractProvider
     public static array $CONFIG_KEY_BLOCK_LIST = [
         // shop instance, license related or consent data
         'core.app.shopId',
+        'core.app.shopIdV2',
```
MalteJanz, line 29: Just curious: I'm also wondering if there might be other keys where a `V2` was added for example and we didn't notice yet.

I think there is also a decorator / some filter mechanism in Rufus / SaaS repos to not allow writes or reads for certain keys but I couldn't find the class quickly. Might be worth looking into again and comparing that against our list here, we should also add a comment where these files can be found with their lists to block.

MalteJanz (approved):
LGTM 👍

DennisGarding (approved):

src/DataProvider/Provider/Data/SystemConfigProvider.php
jozsefdamokos: I found `FilteringSystemConfigSubscriber` in rufus. I think all of them are covered already by the migration plugin because this subscriber was not updated for years. Im not sure maybe we missed updating it because `shopIdV2` is missing from there. Should we add it?

larskemper (approved):

src/DataProvider/Provider/Data/SystemConfigProvider.php
MalteJanz: ~~I mean if `core.app.shopId` wasn't in there, and it looks like it wasn't:~~
https://github.com/shopware/Rufus/blob/10509a6bb1ad0c60255c2e92345c3e74c98100a5/src/Core/System/SystemConfig/Api/FilteringSystemConfigSubscriber.php#L22

~~I don't see why we should add it there now.~~ Nevermind it's in there, we could give @shopware/product-cc-framework / @AydinHassan a heads up, that the SaaS specific block list of `system_config` keys for reading / writing should be checked again for it's up to date ness, especially because of the added app and services stuff the last years. Do you guys want a issue for this? 🙂 

And we have a little different use case, we don't want to migrate anything that is either a "secret" to SaaS or that should be uniquely set up for each Shopware instance

ennasus4sun (approved):

src/DataProvider/Provider/Data/SystemConfigProvider.php
AydinHassan: Might be worth a message in `#product-cc-framework`

PR #93, opened 2025-11-26
feat: prevent multiple connections to the same system

larskemper:
resolves https://github.com/shopware/shopware/issues/13670

SwagMigrationConnector pr: https://github.com/shopware/SwagMigrationConnector/pull/5/

This pull request introduces a fingerprinting mechanism for migration connections to prevent duplicate source system connections.

##### Shopware 6

For Shopware 6 connections, we use the `shopIdV2` property from `system_config` as the connection fingerprint. This ID uniquely identifies every Shopware 6 shop. 

##### Shopware 5

Since Shopware 5 doesn't have a shopId and the SBP identifies shops by domain, we extract `esdKey` and `installationDate` from `s_core_config_elements`, hash them together, and use the resulting hash as the connection fingerprint: 

- `esdKey`: An automatically generated identifier for Shopware 5 shops, present after initial installation. In Shopware 5, it's used to generate serial numbers for digital products. (Note: Theoretically editable by the shop owner, but this is highly unlikely in practice.) 
- `installationDate`: The datetime set during the Shopware 5 installation process.

=> we can of course use other fields and hash them if you have a idea for more appropriate ones


src/Migration/Connection/ConnectionFingerprintService.php
```diff
@@ -0,0 +1,153 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\NotFilter;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Util\Hasher;
+use SwagMigrationAssistant\Profile\Shopware\Gateway\Api\ShopwareApiGateway;
+use SwagMigrationAssistant\Profile\Shopware\Gateway\Local\ShopwareLocalGateway;
+
+#[Package('fundamentals@after-sales')]
+class ConnectionFingerprintService
```
MalteJanz: I like the idea of "fingerprinting" the source system but to be honest I don't see any value in how it's implemented here ✌️ . I'm missing a clear concept on how fingerprinting is done to detect the same source system regardless of the used gateway or credentials.

In this implementation it's basically only hashing the gateway + credentials, which is mostly useless. Because you can still create connection with a different gateway or for example a different URL that is pointing to the same system, and this implementation wouldn't detect it.

MalteJanz (changes requested):

src/Migration/Connection/ConnectionFingerprintService.php
larskemper: mhh, ok and do you have an idea how to improve it?

What I wanted to achieve here is that the user does not create a connection with the same credentials, so more like a easy validation, not that bulletproof for sure.

Another thing we could try is to generating some kind of uuid for the system when the plugin is installed or generate some hash and expose it via api and include it here, but that is probably a backward compatibility problem.

MalteJanz: > What I wanted to achieve here is that the user does not create a connection with the same credentials, so more like a easy validation, not that bulletproof for sure.

I see, but I don't think that is a common problem in practice (e.g. in reality you mostly have one connection, so why would you create the same again). At least in the past we had some cases where people started for example with a local connection and later created another one for API (or the other way around). These cases are much more problematic and are more worth looking into.

> mhh, ok and do you have an idea how to improve it?

Sure, as you already mentioned we need some kind of identifier that uniquely identifies each source system. For SW6 we could add and read the `shopIdV2` `system_config` entry to the environment information and save that as it's fingerprint. It's used for the app system and should uniquely identify a SW6 installation.

For SW5 we would need to find something similar, on top of my head I don't know what yet, so we have to research a bit. Might also be worth looking into how the `shopIdV2` in SW6 is created and if we could apply some of the same logic for SW5 🤔 

For Magento, I have no idea. But that also shouldn't have the highest priority and if there are no fingerprint checks for Magento, that is also fine for now (means empty fingerprint `!==` empty fingerprint to still allow multiple connection if no fingerprint is available)

jozsefdamokos: I agree that we should try to rule out using local and API connections to the same shop as that sounds more like a more common real life scenario. For SW6 we are fine using `shopIdV2`, for SW5 Im also not sure what we can use, if there is not a similar thing we could use a combination of identifiers... I guess we will have to update the connector plugin for SW5 to expose that then?

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -168,7 +168,8 @@
             "SWAG_MIGRATION__DATABASE_CONNECTION_ERROR": "Die Datenbank Verbindung konnte nicht hergestellt werden.",
             "SWAG_MIGRATION__DATABASE_CONNECTION_ATTRIBUTES_WRONG": "Die Datenbankverbindung hat nicht die richtigen Attribute und diese können nicht gesetzt werden.",
             "SWAG_MIGRATION__IS_RUNNING": "Eine Migration ist zurzeit im Gange. Du kannst deshalb die Zugangsdaten nicht bearbeiten, bis die Migration abgeschlossen ist.",
-            "SWAG_MIGRATION__SSL_REQUIRED": "Wir haben festgestellt, dass die angegebene Shop-Domain eine SSL-Verbindung erfordert."
+            "SWAG_MIGRATION__SSL_REQUIRED": "Wir haben festgestellt, dass die angegebene Shop-Domain eine SSL-Verbindung erfordert.",
+            "SWAG_MIGRATION__DUPLICATE_SOURCE_CONNECTION": "Es besteht bereits eine Verbindung mit denselben Zugangsdaten. Bitte verwenden Sie eine andere Verbindung oder bearbeiten Sie die bestehende."
```
ennasus4sun:
```suggestion
            "SWAG_MIGRATION__DUPLICATE_SOURCE_CONNECTION": "Es besteht bereits eine Verbindung mit denselben Zugangsdaten. Bitte verwenden Sie eine andere Verbindung oder bearbeiten Sie die Bestehende."
```

tests/Migration/Services/RunServiceTest.php
```diff
@@ -136,6 +141,154 @@ protected function setUp(): void
...
+    public function testUpdateConnectionCredentialsWhenMigrationIsRunning(): void
+    {
+        $run = new SwagMigrationRunEntity();
+        $run->setId(Uuid::randomHex());
+
+        $this->runRepo = new StaticEntityRepository([
+            new SwagMigrationRunCollection([$run]),
+        ], new SwagMigrationRunDefinition());
+
+        static::expectException(MigrationException::class);
```
ennasus4sun: Sometimes I had read to better use this method `expectExceptionObject()`:

```suggestion
        static::expectExceptionObject(MigrationException::migrationIsAlreadyRunning());
```

tests/unit/Exception/MigrationExceptionTest.php
```diff
@@ -0,0 +1,30 @@
...
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Exception\MigrationException;
+
+/**
+ * @internal
+ */
+#[Package('fundamentals@after-sales')]
+#[CoversClass(MigrationException::class)]
+class MigrationExceptionTest extends TestCase
```
ennasus4sun: do we need to test every exception?
we could add `@codeCoverageIgnore` to src/Exception/MigrationException.php
This is also used from time to time in other exception classes like https://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/Cart/CartException.php#L23

larskemper: some domains do, some domains do not. Does not cost much time and prevents unintended changes, so why not 😅

ennasus4sun: yes, was just an idea... To me, this doesn't make much sense...

MalteJanz: I would also vote for `@codeCoverageIgnore` on the `Exception` class and not writing a test for it.

In the end it's mostly just static data and your tests duplicates that to check it's the same static data. It's not testing any behavior, so it doesn't provide much value.

> Does not cost much time and prevents unintended changes, so why not
- It does cost us time to write, read and maintain
- Yes it will fail if you adjust any of the static data, but you would only do that by intention anyways and then it's just extra work for fixing the test. It's not testing for anything that might change unintentionally, like a slight change in behavior.
- In the end this test is just there to keep the code coverage up, which shouldn't be the only goal of a test

src/Migration/Run/RunService.php
```diff
@@ -134,19 +136,42 @@ public function getRunStatus(Context $context): MigrationState
...
      */
     public function updateConnectionCredentials(Context $context, string $connectionUuid, ?array $credentialFields): void
     {
         if ($this->isMigrationRunning($context)) {
             throw MigrationException::migrationIsAlreadyRunning();
         }
 
-        $context->scope(MigrationContext::SOURCE_CONTEXT, function (Context $context) use ($connectionUuid, $credentialFields): void {
+        /** @var SwagMigrationConnectionEntity $connection */
+        $connection = $this->connectionRepo->search(new Criteria([$connectionUuid]), $context)->first();
```
ennasus4sun:
```suggestion
        $connection = $this->connectionRepo->search(new Criteria([$connectionUuid]), $context)->getEntities()->first();
```
could be that you can remove the `@var` annotation with this

ennasus4sun (approved):

jozsefdamokos (approved):

DennisGarding (approved):

src/Migration/Run/RunService.php
```diff
@@ -134,19 +136,42 @@ public function getRunStatus(Context $context): MigrationState
...
+        );
+
+        $hasDuplicate = $this->connectionFingerprintService->check(
+            $fingerprint,
+            $context,
+            $connection->getId(),
+        );
+
+        if ($hasDuplicate) {
+            throw MigrationException::duplicateSourceConnection();
```
vintagesucks: Doesn't doing all of this in `updateConnectionCredentials` only prevent the already created duplicate connection from receiving a fingerprint?

I tested this locally (SW6 → SW6) and creating a duplicate connection currently behaves like this:
- While entering the details for the second connection, an empty connection is already created in `swag_migration_connection`
- After submitting the duplicate details, I do see the new "A connection to the same source system already exists." warning, but the connection is updated nonetheless, and the `swag_migration_connection` row now contains the credentials, only the fingerprint is missing:
  
  image
- After dismissing the modal with "Close", the duplicate connection is not only created, but also active

Image

vintagesucks (changes requested):

src/Migration/Run/RunService.php
larskemper: That's a general migration design issue we didn't refactor until now. I suspect it was done this way because the wizard is url based: each step lives on its "own page", so state isn't shared and nothing gets stored in vuex. That said, this shouldn't prevent existing duplicate connections from receiving a fingerprint, the service only checks for existing fingerprints and exits early when none are present. So the first connection should get a fingerprint and is valid, the second not...

but in this case it seems different, I will look into that, probably need to add a filter for null in the service 👍

vintagesucks: What I don't get is what this PR achieves then, because the only difference to me as a user of this is that I see the warning modal, but nothing "prevent[s] multiple connections to the same system" - what am I missing here? 😅

larskemper: i get what you mean now, it validates newly created connections: "prevent **creating** multiple connections to the same system". Existing connections do not have a fingerprint, that's true, so their are not validated

larskemper: found the issue: we update the connection first and then did a connection check, so creds where written before checking. I refactored it and now we first check the connection with given creds and then update the entity & generate the fingerprint

vintagesucks: Better 👍 but I still think there's something strange about it 🤔

---

1. UI briefly shows connection success before the warning:

https://github.com/user-attachments/assets/bcb61e1f-1311-476d-8348-b9729656e938

---

2. As a user, I am still left with an empty connection when this happens (now without credentials), which is a confusing state to me. While creating an exact duplicate of my existing one was prevented, it would be nicer if this empty connection would either be cleaned up or not created at all, so I am back to my initial (working) connection when I exit this modal, instead of seeing this:

image

---

Feel free to move this to follow-up issues due to how all of this works, but I think this feature does not make much sense with the behavior described in 2.

larskemper: yeah, definitely strange but at least "not connected" 😅. Let's create a follow up bug ticket for that 👍

vintagesucks: Do you want to create the ticket(s)?

vintagesucks (approved):

src/Migration/Run/RunService.php
larskemper: https://github.com/shopware/SwagMigrationAssistant/pull/93#discussion_r2613878516

src/Migration/Connection/Fingerprint/MigrationFingerprintService.php
```diff
@@ -0,0 +1,75 @@
...
+                } catch (\Throwable) {
+                    return null;
+                }
+            }
+        }
+
+        return null;
+    }
+
+    public function check(?string $fingerprint, Context $context, ?string $excludeConnectionId): bool
```
MalteJanz: Nitpick: A doc comment would be helpful here, because just the method name `check` doesn't say much about the returned boolean, e.g. `returns true if a connection with the same fingerprint already exists`

src/Controller/StatusController.php
```diff
@@ -247,6 +247,12 @@ public function checkConnection(Request $request, Context $context): JsonRespons
...
         }
 
+        $credentialFields = $request->request->all('credentialFields');
+
+        if (!empty($credentialFields)) {
+            $connection->setCredentialFields($credentialFields);
+        }
+
         $migrationContext = $this->migrationContextFactory->createByConnection($connection);
         $information = $this->migrationDataFetcher->getEnvironmentInformation($migrationContext, $context);
```
MalteJanz, line 261: Because retrieving the environment data from the source system can be quite expensive, I would suggest adding the fingerprinting here in this `check-connection` endpoint, where the `$information` is already available

You can then write it back to the connection DB table if it is still `null` there, so it only gets inserted once

src/Migration/Connection/Fingerprint/Provider/Shopware5FingerprintProvider.php
```diff
@@ -0,0 +1,118 @@
...
+        $connection->setCredentialFields($credentialFields);
+        $migrationContext = $this->migrationContextFactory->createByConnection($connection);
+
+        $environmentReader = $this->getEnvironmentReader($connection->getGatewayName());
+
+        if ($environmentReader === null) {
+            return null;
+        }
+
+        $response = $environmentReader->read($migrationContext);
```
MalteJanz: As written in my other comment, this call can be quite expensive and ideally should only happen once during the whole connection setup / on initial page load before starting a migration. With this change it likely happens twice, first on saving the connection credentials, and then again by the `check-connection` endpoint, which isn't great if both might take e.g. 5 seconds in the real world.

I would suggest leaving this out of here, and just processing a given `$information` from an environment call in `check-connection` endpoint, passing that into here.

Same applies for all other fingerprint providers as well.

src/Migration/Connection/Fingerprint/Provider/Shopware5FingerprintProvider.php
```diff
@@ -0,0 +1,118 @@
...
+
+    private function getEnvironmentReader(string $gatewayName): ?EnvironmentReaderInterface
+    {
+        if ($gatewayName === ShopwareApiGateway::GATEWAY_NAME) {
+            return $this->apiEnvironmentReader;
+        }
+
+        if ($gatewayName === ShopwareLocalGateway::GATEWAY_NAME) {
+            return $this->localEnvironmentReader;
+        }
```
MalteJanz: To be honest, you could even add the extra bit of logic for generating a fingerprint and storing that on the connection to each individual "environmentReader", which is already specific for a given source system + gateway combination.

Then we don't need all these extra interface + individual services, except the `MigrationFingerprintService` itself for comparing it (you might even call it in the "environmentReader"s to generate with some utility methods)

larskemper: will refactor then, third time's a charm 😄

MalteJanz: Sorry 😄 ✌️

src/Migration/Run/RunService.php
```diff
@@ -134,18 +134,27 @@ public function getRunStatus(Context $context): MigrationState
...
         }
 
-        $context->scope(MigrationContext::SOURCE_CONTEXT, function (Context $context) use ($connectionUuid, $credentialFields): void {
+        $connection = $this->connectionRepo->search(
+            new Criteria([$connectionUuid]),
+            $context,
+        )->getEntities()->first();
+
+        if ($connection === null) {
+            throw MigrationException::noConnectionFound();
```
MalteJanz: Why is it needed to first fetch the connection before updating it? Just to throw this exception? In that case you could also add a try / catch block around the update 🤔

src/Resources/app/administration/src/module/swag-migration/store/migration.store.ts
```diff
@@ -364,8 +364,22 @@ const migrationStore = Shopware.Store.register({
...
+                const code = error?.response?.data?.errors[0]?.code;
+
+                if (!code) {
+                    await this.createErrorNotification('swag-migration.api-error.checkConnection');
+                    return;
+                }
+
+                const errorMessageSnippet = `swag-migration.wizard.pages.credentials.error.${code}`;
+
+                if (!Shopware.Snippet.tc(errorMessageSnippet)) {
```
MalteJanz: This if branch will never be executed (it should always return a string, even if the it's just the key if it doesn't exists). I guess you want to check if the snippet exists?

```suggestion
                if (!Shopware.Snippet.te(errorMessageSnippet)) {
```

tests/Migration/Controller/StatusControllerTest.php
```diff
@@ -690,4 +740,85 @@ private function jsonResponseToArray(?Response $response): array
...
+                    $environmentInformation->getAdditionalData(),
+                    $environmentInformation->getRequestStatus(),
+                    $environmentInformation->isMigrationDisabled(),
+                    $environmentInformation->getDisplayWarnings(),
+                    $environmentInformation->getTargetSystemCurrency(),
+                    $environmentInformation->getSourceSystemCurrency(),
+                    $environmentInformation->getSourceSystemLocale(),
+                    $environmentInformation->getTargetSystemLocale(),
+                    $fingerprint
+                );
```
MalteJanz: nitpick: what speaks against adding a setter for the `$fingerprint` instead of calling the constructor again?

MalteJanz (approved):
Besides my comments it looks great now. Thanks for refactoring it another time 💪

src/Migration/Run/RunService.php
larskemper: missed to delete that

tests/Migration/Controller/StatusControllerTest.php
larskemper: didn't wanted to change api for a test, but yes afterwards i also would have created setter 😅

MalteJanz (approved):

PR #94, opened 2025-11-26
chore: check naming consistency across error resolution

larskemper:
resolves https://github.com/shopware/shopware/issues/13671

#### Changes:

- package annotation `after-sales` => `fundamentals@after-sales`
- `MigrationFix` services => `ErrorResolution` services
- moved all relevant backend classes to the same domain `ErrorResolution`
- renamed migration step `apply-fixes` => `error-resolution`
- renamed non entity classes with `SwagMigration` prefix to use `Migration` prefix


ennasus4sun (approved):

src/Migration/ErrorResolution/SwagMigrationErrorResolution.php
```diff
@@ -5,14 +5,14 @@
...
+namespace SwagMigrationAssistant\Migration\ErrorResolution;
 
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
 use SwagMigrationAssistant\Exception\MigrationException;
 
-#[Package('after-sales')]
-class MigrationFix
+#[Package('fundamentals@after-sales')]
+class SwagMigrationErrorResolution
```
MalteJanz: I would leave this as `MigrationFix`, because it's based on the `SwagMigrationFixEntity` and related DB table.

I think a `ErrorResolutionService` that uses this `MigrationFix`es makes sense. Otherwise you would need to rename everything fix related, all the way down to DB level.

src/Migration/ErrorResolution/SwagMigrationErrorResolutionService.php
```diff
@@ -5,15 +5,15 @@
...
 
 use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
 use Shopware\Core\Framework\Log\Package;
 use Shopware\Core\Framework\Uuid\Uuid;
 
-#[Package('after-sales')]
-class MigrationFixApplier
+#[Package('fundamentals@after-sales')]
+class SwagMigrationErrorResolutionService
```
MalteJanz: I wouldn't add `Swag` as a prefix here. It's only used for the Entity classes, because there it mirrors the DB table name, which itself is prefixed by `swag_` to prevent collisions with core or other plugin tables.

For our services this isn't needed, they are already namespaced. We still often have the `Migration` prefix to make it more clear that this is a thing that applies specifically to migration, and not something in the core (e.g. `ErrorResolutionService` sounds very generic, what errors?) 
```suggestion
class MigrationErrorResolutionService
```

src/Migration/Mapping/MappingServiceInterface.php
```diff
@@ -84,7 +84,7 @@ public function writeMapping(): void;
      */
     public function getMappings(string $connectionId, string $entityName, array $ids, Context $context): EntitySearchResult;
 
-    public function hasValidMappingByEntityUuid(string $connectionId, string $entityName, string $entityId, Context $context): bool;
+    public function hasValidMappingByEntityId(string $connectionId, string $entityName, string $entityId, Context $context): bool;
```
MalteJanz, line 87: Do you already want to document this break?

MalteJanz (approved):
After my comments are resolved, it LGTM 👍

src/Migration/ErrorResolution/SwagMigrationErrorResolutionService.php
larskemper: yeah, wasn't sure, cause I also saw a Service having the SwagMigration prefix, but I will have a look and maybe also rename them too

larskemper: found some non Entity classes from Logging & Validation that had the SwagMigration prefix, renamed them to use just the Migration prefix

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -106,7 +106,7 @@ public function writeData(MigrationContextInterface $migrationContext, Context $
         }
 
         $convertedValues = array_values($converted);
-        $this->fixApplier->apply($convertedValues, $migrationContext->getConnection()->getId(), $migrationContext->getRunUuid());
+        $this->errorResolutionService->apply($convertedValues, $migrationContext->getConnection()->getId(), $migrationContext->getRunUuid());
```
jozsefdamokos: Since this is not yet named `fixApplier` should we rename the method to `applyFixes`? `errorResolutionService->apply` sounds incomplete now

jozsefdamokos (approved):

UPGRADE.md
```diff
@@ -1,5 +1,8 @@
 # 16.0.0
 
+- [BREAKING] [#94](https://github.com/shopware/SwagMigrationAssistant/pull/94) - chore!: check naming consistency across error resolution
+  - [BREAKING] Renamed method `hasValidMappingByEntityUuid()` to ``hasValidMappingByEntityId() in `SwagMigrationAssistant\Migration\Mapping\MappingServiceInterface` and implementation `SwagMigrationAssistant\Migration\Mapping\MappingService`
```
MalteJanz: I guess you could now also mention the renaming of `MigrationLogBuilder` here, I guess that class was public? 🤔

MalteJanz (approved):

src/Migration/Service/MigrationDataWriter.php
larskemper: sounds good 👍

UPGRADE.md
larskemper: the class is part of the error resolution feature, so unreleased 😅

larskemper: i will have a look which classes are public

PR #98, opened 2025-12-04
fix: change connection modal

vintagesucks:
The change connection modal is currently broken:

image

### DIFF
--- a/src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard-page-connection-select/index.ts
+++ b/src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard-page-connection-select/index.ts
@@ -86,8 +86,13 @@ export default Shopware.Component.wrapComponentConfig({
             });
 
             this.selectedConnectionId = newId;
-            this.$emit('onChildRouteReadyChanged', !!connection);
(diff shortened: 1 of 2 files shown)


src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard-page-connection-select/index.ts
```diff
@@ -86,8 +86,13 @@ export default Shopware.Component.wrapComponentConfig({
             });
 
             this.selectedConnectionId = newId;
-            this.$emit('onChildRouteReadyChanged', !!connection);
-            this.$emit('onConnectionSelected', connection);
+
+            if (connection) {
```
larskemper: nice catch, but it’s usually helpful to emit null changes as well. In this case the root cause appears to be in the parent component `swag-migration-wizard`: the template tries to access the connection before it’s loaded/selected. So we could just add chaining when passing the prop:

```html
<component
    ...
    :current-connection-id="connection?.id"
    :credentials="connection?.credentialFields"
    ...
/>
```

larskemper (changes requested):

larskemper (approved):

ennasus4sun (approved):

PR #100, opened 2025-12-08
refactor: unify public api

larskemper:
resolves https://github.com/shopware/shopware/issues/13742

#### Changes

- mark services internal & dispatch events for extensibility
- mark classes we want public, but not extended final


src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
```diff
@@ -125,9 +126,7 @@ public function withExceptionTrace(array $exceptionTrace): self
      */
     public function build(string $logClass): AbstractMigrationLogEntry
     {
-        if (!class_exists($logClass) || !is_subclass_of($logClass, AbstractMigrationLogEntry::class)) {
-            throw MigrationException::failedToCreateMigrationLog($logClass);
```
jozsefdamokos, line 129: Isnt it better to keep the more specific exception here?

larskemper: not really necessary i think. Its obviously an dev config issue that will not happen in prod or hard fail anyways

src/Migration/Validation/MigrationValidationService.php
```diff
@@ -33,6 +33,9 @@
 use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationUnexpectedFieldLog;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
+/**
+ * @internal
```
jozsefdamokos, line 37: I can see this one being extended by third parties

jozsefdamokos (approved):

src/Migration/Validation/MigrationValidationService.php
larskemper: they can with the dispatched events in the validator. Also I think validation should be our thing and not touched so much in extentions

ennasus4sun (approved):

src/Controller/HistoryController.php
```diff
@@ -28,7 +28,7 @@ class HistoryController extends AbstractController
 {
     public function __construct(
```
MalteJanz, line 32: Just a minor thing while you are at it: Would be nice if all our `__construct` functions of our Controllers would be marked `@internal`, so we can easily change the parameters without breaks.

src/Migration/ErrorResolution/Event/MigrationPostErrorResolutionEvent.php
```diff
@@ -0,0 +1,35 @@
...
+use SwagMigrationAssistant\Migration\ErrorResolution\MigrationErrorResolutionContext;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @final
+ */
+#[Package('fundamentals@after-sales')]
+class MigrationPostErrorResolutionEvent extends Event implements ShopwareEvent
+{
+    public function __construct(private readonly MigrationErrorResolutionContext $errorResolutionContext)
```
MalteJanz, line 27: I guess third-parties don't need to call the constructor on event classes right? In that case also mark just this constructor `@internal`, so we can easily add more data in the future

src/Migration/ErrorResolution/Event/MigrationPreErrorResolutionEvent.php
```diff
@@ -0,0 +1,35 @@
...
+use SwagMigrationAssistant\Migration\ErrorResolution\MigrationErrorResolutionContext;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @final
+ */
+#[Package('fundamentals@after-sales')]
+class MigrationPreErrorResolutionEvent extends Event implements ShopwareEvent
+{
+    public function __construct(private readonly MigrationErrorResolutionContext $errorResolutionContext)
```
MalteJanz, line 27: see above

src/Migration/ErrorResolution/MigrationErrorResolutionContext.php
```diff
@@ -0,0 +1,78 @@
...
+ * @final
+ */
+#[Package('fundamentals@after-sales')]
+class MigrationErrorResolutionContext
+{
+    /**
+     * @param array<int|string, array<int|string, mixed>> $data
+     * @param array<string, list<MigrationFix>> $fixes
+     */
+    public function __construct(
```
MalteJanz, line 27: see above, in general if the class is already marked with `@final` (so no inheritance from it / decoration that might need to call the constructor) and if it is only constructed by us / the dependency injection (DI) container, then I would also always add `@internal` to the constructor just to be explicit that the constructor can change.
This is not need if the whole class is already marked `@internal`

src/Migration/Validation/Event/MigrationPostValidationEvent.php
MalteJanz, line 27: see above

src/Migration/Validation/Event/MigrationPreValidationEvent.php
MalteJanz, line 27: see above

src/Migration/Validation/MigrationValidationContext.php
MalteJanz, line 31: see above

tests/unit/Migration/ErrorResolution/Event/MigrationPostErrorResolutionEventTest.php
```diff
@@ -0,0 +1,42 @@
...
+use SwagMigrationAssistant\Migration\ErrorResolution\MigrationErrorResolutionContext;
+
+/**
+ * @internal
+ */
+#[Package('fundamentals@after-sales')]
+#[CoversClass(MigrationPostErrorResolutionEvent::class)]
+class MigrationPostErrorResolutionEventTest extends TestCase
+{
+    public function testPostErrorResolutionEvent(): void
```
MalteJanz: What value does this test provide? The event itself is just a dump data object, similar to our exception objects. I would just mark the event itself `@codeCoverageIgnore` 🤔

A much more useful test would be if this event actually gets emitted in the correct place, which you have already done 🙂

tests/unit/Migration/ErrorResolution/Event/MigrationPreErrorResolutionEventTest.php
```diff
@@ -0,0 +1,42 @@
...
+use SwagMigrationAssistant\Migration\ErrorResolution\MigrationErrorResolutionContext;
+
+/**
+ * @internal
+ */
+#[Package('fundamentals@after-sales')]
+#[CoversClass(MigrationPreErrorResolutionEvent::class)]
+class MigrationPreErrorResolutionEventTest extends TestCase
+{
+    public function testPreErrorResolutionEvent(): void
```
MalteJanz: see above

tests/unit/Migration/ErrorResolution/MigrationErrorResolutionContextTest.php
```diff
@@ -0,0 +1,63 @@
...
+use SwagMigrationAssistant\Migration\ErrorResolution\MigrationErrorResolutionContext;
+
+/**
+ * @internal
+ */
+#[Package('fundamentals@after-sales')]
+#[CoversClass(MigrationErrorResolutionContext::class)]
+class MigrationErrorResolutionContextTest extends TestCase
+{
+    public function testErrorResolutionContext(): void
```
MalteJanz: Same here, if this context would have special logic in its getters / setters, I would say it's totally fine to test it. But I don't see anything special in that class under test

MalteJanz (approved):
Mostly some nitpicks / further suggestions of mine. Overall looks good already 👍

src/Migration/ErrorResolution/Event/MigrationPostErrorResolutionEvent.php
larskemper: true, their just need to access the getters 👍

src/Migration/ErrorResolution/MigrationErrorResolutionContext.php
larskemper: the class is marked `@final` so thrird party can use but not extend. I will mark the constructor `@internal` 👍

PR #102, opened 2025-12-15
refactor: use domain exception

larskemper:
#### Changes:

- remove unused exceptions
- move all existing exceptions to domain class
- change incorrect status codes
- unify exceptions with same meaning
- jest: adjustment of the config was needed to prevent jest ci from failing (lodash wasn't correctly resolved, coppied fix from platform: [ref](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/jest.config.js#L91))


src/Resources/app/administration/src/module/swag-migration/snippet/de.json
```diff
@@ -161,7 +161,6 @@
             "SWAG_MIGRATION__API_CONNECTION_ERROR__HINT": "Wenn beim Herstellen einer Verbindung Probleme auftreten, lohnt es sich möglicherweise, die aktuell auf Ihrem Server installierte Curl-Version zu überprüfen und zu aktualisieren.",
             "SWAG_MIGRATION__GATEWAY_READ": "Es konnte keine Verbindung zum angegebenen Server hergestellt werden. Bitte überprüfe die angegebene Shop-Domain.",
             "SWAG_MIGRATION__GATEWAY_READ__HINT": "Wenn beim Herstellen einer Verbindung Probleme auftreten, lohnt es sich möglicherweise, die aktuell auf Ihrem Server installierte Curl-Version zu überprüfen und zu aktualisieren.",
-            "SWAG_MIGRATION__INVALID_CONNECTION_AUTHENTICATION": "Deine Zugangsdaten sind nicht korrekt. Bitte überprüfe sie und versuche es erneut.",
```
ennasus4sun, line 164: short question: do we still provide a message when the credentials are not correct?

ennasus4sun (approved):

src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
```diff
@@ -85,23 +84,16 @@ public function __invoke(ProcessMediaMessage $message): void
             $processor = $this->mediaFileProcessorRegistry->getProcessor($migrationContext);
             $workload = $processor->process($migrationContext, $context, $workload);
             $this->processFailures($context, $migrationContext, $processor, $workload);
-        } catch (NoConnectionFoundException $exception) {
+        } catch (MigrationException|\Exception $exception) {
+            $logClass = $exception instanceof MigrationException
+                ? ProcessorNotFoundLog::class : ExceptionRunLog::class;
```
MalteJanz: I don't think this was a good change and a dedicated exception that can be catched individually makes more sense here.

E.g. now with this implementation what happens if the `$processor->process` call also throws a `MigrationException`, because it failed to process some media. Then you would just log that as "processor not found" with a confusing message, which is strange. To be fair, I can't immediately say if that call is actually able to throw a `MigrationException`, but still if not today, it might do that tomorrow 😅

src/Migration/Service/MediaFileProcessorService.php
```diff
@@ -45,8 +45,10 @@ public function processMediaFiles(MigrationContextInterface $migrationContext, C
             if ($currentDataSet === null) {
                 try {
                     $currentDataSet = $this->dataSetRegistry->getDataSet($migrationContext, $mediaFile['entity']);
-                } catch (DataSetNotFoundException $exception) {
-                    $this->logDataSetNotFoundException($migrationContext, $exception);
+                } catch (MigrationException $exception) {
+                    if ($exception->getErrorCode() === MigrationException::DATASET_NOT_FOUND) {
```
MalteJanz: Same applies to this change, but slightly differently:

Previously every other Exception that wasn't `DataSetNotFoundException` was still thrown / bubble up. Now all kinds of `MigrationException` that don't have that specific error code are silently ignored, which is very bad. Sure you could add an else branch and throw it again, but it's just isn't as clean as a proper dedicated exception.

In general every time we explicitly need to catch an Exception, it totally makes sense to have a dedicated Exception class (which could still inherit from the Domain Exception). Btw, that is also documented here 🙂 : https://developer.shopware.com/docs/resources/references/adr/2022-02-24-domain-exceptions.html

src/Migration/Service/MediaFileProcessorService.php
```diff
@@ -59,8 +61,10 @@ public function processMediaFiles(MigrationContextInterface $migrationContext, C
                     $messageMediaUuids = [];
                     $currentCount = 0;
                     $currentDataSet = $this->dataSetRegistry->getDataSet($migrationContext, $mediaFile['entity']);
-                } catch (DataSetNotFoundException $exception) {
-                    $this->logDataSetNotFoundException($migrationContext, $exception);
+                } catch (MigrationException $exception) {
+                    if ($exception->getErrorCode() === MigrationException::DATASET_NOT_FOUND) {
```
MalteJanz: see above

src/Resources/app/administration/src/module/swag-migration/snippet/de.json
larskemper: yes: the one directly below 😄

src/Migration/Service/MediaFileProcessorService.php
larskemper: The function called here can really only throw this MigrationException and depth is also just one layer, so we could also just get rid of the if check and log if a exception is thrown... don't think that with this less complexity a own Exception class is necessary

src/Migration/Service/MediaFileProcessorService.php
larskemper: same here

src/Migration/MessageQueue/Handler/ProcessMediaHandler.php
larskemper: don't no why I though this was good 😅. Actually the solution here I think is to not have every thing in the same try-catch. The Registry can throw a MigrationException, the rest all kind ...

vintagesucks (approved):

src/Migration/Service/MediaFileProcessorService.php
MalteJanz: Even if that is the case today, I still think it's a code smell, because it's very easy to introduce a bug (with your updated version emitting a log that is misleading).

For example if someone is working on the class in the try block, for whatever reason throwing another type of `MigrationException`, he can't safely assume that it will be properly handled in all places and would need to adjust this try block as well / has to know the whole codebase.

If it would be a dedicated exception, it's much clearer.

I'm totally aware that it is very unlikely that someone would change these Registries and suddenly throw a different exception. But on the other hand having multiple try / catch blocks after one another also isn't very clean. I would still vote for leaving a dedicated exception for these cases.

src/Migration/Service/MediaFileProcessorService.php
MalteJanz: same here 😅

src/Migration/Service/MediaFileProcessorService.php
larskemper: but with this argument we can't use Domain Exceptions, actually any Exception or try-catch anywhere. Yeah when somebody changes the Exception in the class "below" where its thrown and not the catch then we have the same situation, also with any other Exception.

MalteJanz: That's not totally true, in other places you might catch our generic domain exception and deal with it like "something gone wrong with the migration", e.g. by creating a proper log entry.

But in this specific case you are only interested in one very specific exception (that the registry can't find something) and are handling it exactly like that, but still catching a generic `MigrationException`, which could be many things.

If you would deal with the `MigrationException` here just like "something gone wrong with the migration" and emitting a generic error log, everything would be totally fine. I just don't like that it's catching a `MigrationException` and then always automatically assumes it must be that the registry couldn't find the item, which might not always be true.

And that is exactly why we still use dedicated exception classes next to our domain exceptions, when they make sense / are catched in specific places.

larskemper: I got that part, but what speaks agains `if ($exception->getErrorCode() === MigrationException::PROCESSOR_NOT_FOUND)` then?

MalteJanz: The missing `else` block 😅. In this specific place an `else { throw $e; }` would have been also fine, it just makes sure the exception isn't "silenced"

Additionally it's not as clean to read as multiple / a more dedicated catch block. But that's only a nitpick

MalteJanz (approved):

src/Migration/Service/MediaFileProcessorService.php
```diff
@@ -41,14 +41,19 @@ public function processMediaFiles(MigrationContextInterface $migrationContext, C
...
         $messageMediaUuids = [];
+
         foreach ($mediaFiles as $mediaFile) {
             if ($currentDataSet === null) {
                 try {
                     $currentDataSet = $this->dataSetRegistry->getDataSet($migrationContext, $mediaFile['entity']);
-                } catch (DataSetNotFoundException $exception) {
-                    $this->logDataSetNotFoundException($migrationContext, $exception);
+                } catch (MigrationException $exception) {
+                    if ($exception->getErrorCode() === MigrationException::DATASET_NOT_FOUND) {
```
mitelg, line 50: for exact that use case we still use specific child domain exception :grimacing: 
those are then simply thrown from the main domain exception class.

PR #105, opened 2025-12-19
refactor: migration validation

larskemper:
#### General

- unified logging and exception handling (own validation exception class)
- split invalid field value in required / optional field log (required is an error, optional a warning)
- added translation log for future fix implementation
- fix conversion of order custom fields (before was set to invalid "null" if not existing)
- fix sizing and coloring of ssl url in ui
- fix unpinned padding of log group table

#### Validation wise

- added check for database field definition: field now get marked as required if: definition has required flag, database field is not nullable and has no default value
- removed check for unexpected fields (adds unnecessary noise to logs and errors)
- removed db lookup of associations ids (n+1 query problem)
- removed validation of association fields by serializer (often false positive, cause of missing write context data)
- added a more naive / basic check for association fields: correct data structure and valid ids


src/Migration/Validation/MigrationValidationService.php
```diff
@@ -37,13 +39,19 @@
  * @internal
  */
 #[Package('fundamentals@after-sales')]
-readonly class MigrationValidationService
+class MigrationValidationService
 {
+    /**
+     * @var array<string, list<string>>
+     */
+    private array $requiredColumnsCache = [];
```
MalteJanz: Because this is used for memoization / a local cache, please also implement the reset interface and add the proper DI container tag: https://developer.shopware.com/docs/resources/references/adr/2022-03-09-reset-class-state-during-requests.html

src/Migration/Validation/MigrationValidationService.php
```diff
@@ -106,68 +114,91 @@ public function validate(
         return $validationContext->getValidationResult();
     }
 
+    /**
+     * Validates that all required fields are present and that no unexpected fields exist.
```
MalteJanz: > and that no unexpected fields exist.

Are unexpected fields actually an issue? e.g. do they cause an exception on write? You could double check that by calling the admin API with data for an entity that includes "unexpected fields" 🤔 

If it works fine I wouldn't report it, because imagining you migrate from SW6.7.5.0 to SW6.7.4.0 (for whatever reason), and because the source system is a minor version ahead, it might include an additional field that doesn't exists in the target system yet. Right now our converters for SW6 are rather dump and just copy the data over, which means it might produce an validation error that isn't an issue in practice

src/Migration/Validation/MigrationValidationService.php
MalteJanz: Just trying to understand: why is this extra `Required` flag needed for running the serializer? A short code comment would be helpful here 🙂

src/Migration/Validation/MigrationValidationService.php
```diff
@@ -240,16 +265,69 @@ private function validateAssociations(MigrationValidationContext $validationCont
             );
```
MalteJanz: Thinking more about this: we might ignore a edge case here:

Yes in the best case, e.g. an product is converted (and has a mapping) before an order is converted and then trying to look up that product id in the mappings DB should work. So far so good.

But in practice we also often convert an entity together with some associations. e.g. the product data contains it's manufacturer data:
https://github.com/shopware/SwagMigrationAssistant/blob/5d4ec50b03152b8433e76cd8eddbb4cf87167905/src/Profile/Shopware/Converter/ProductConverter.php#L352

which still creates a mapping entry:
https://github.com/shopware/SwagMigrationAssistant/blob/5d4ec50b03152b8433e76cd8eddbb4cf87167905/src/Profile/Shopware/Converter/ProductConverter.php#L696-L701

but that mapping entry isn't immediately flushed to the DB, only after a whole batch is processed:
https://github.com/shopware/SwagMigrationAssistant/blob/5d4ec50b03152b8433e76cd8eddbb4cf87167905/src/Migration/Service/MigrationDataConverter.php#L71

The mapping service correctly handles this with it's internal cache:
https://github.com/shopware/SwagMigrationAssistant/blob/8f780ffb229c75e7a8f59103115be74ae57de60b/src/Migration/Mapping/MappingService.php#L114-L117

So doing the same mapping lookup twice will return the same uuid instead of generating a new one.

BUT it looks like your `MigrationValidationService` is called once per entity before the mappings are flushed to the DB:
https://github.com/shopware/SwagMigrationAssistant/blob/5d4ec50b03152b8433e76cd8eddbb4cf87167905/src/Migration/Service/MigrationDataConverter.php#L114-L120

Which isn't a problem in itself, but here comes the

TLDR: the marked function call to `$this->mappingService->hasValidMappingByEntityId` is implemented like this:
https://github.com/shopware/SwagMigrationAssistant/blob/8f780ffb229c75e7a8f59103115be74ae57de60b/src/Migration/Mapping/MappingService.php#L217-L231

And thus always reaches for the DB, without respecting the local cache first. So these cases likely create validation errors, even though the mappings exists, just not yet in the DB 🤯 

=> that `hasValidMappingByEntityId` method needs to be adjusted to search through the `$this->writeArray` first (unfortunately it can't use the `$this-mappings` cache, because that is indexed by old identifiers, not the new entityId)

MalteJanz (changes requested):

src/Migration/Validation/MigrationValidationService.php
larskemper: ntk 👍

src/Migration/Validation/MigrationValidationService.php
larskemper: the required flag controls the flow in `AbstractFieldSerializer` ([ref](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/FieldSerializer/AbstractFieldSerializer.php#L111)). Without the serializer skips field validation.

will add a comment

src/Migration/Validation/MigrationValidationService.php
MalteJanz: Also thinking further about it: Is this kind of validation really necessary in practice? Which issues in a migration does it detect?

Because it creates another [`N+1 Query` problem](https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it) by potentially having to look for multiple existing mappings in the DB for every processed entity, which can't really be preloaded with the whole entity batch.

And that adds more complexity and brings us further away from solving that "N+1 Query Problem" for the migration in general, thus slower migrations / worse performance

src/Migration/Validation/MigrationValidationService.php
larskemper: if not specifically implemented it will not throw by default, but the error is also just a warning + unfixable

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -170,6 +174,96 @@ public function testShouldValidateStructureAndFieldsValues(array $convertedData,
...
+            'cmsPageVersionId', // has default value, but has required flag
+        ];
+
+        static::assertCount(
+            0,
+            \array_intersect($expectedMissingFields, $missingFields)
+        );
+    }
+
+    public function testShouldLogWhenEntityHasNowId(): void
```
ennasus4sun:
```suggestion
    public function testShouldLogWhenEntityHasNoId(): void
```
?

ennasus4sun (approved):

jozsefdamokos (approved):

src/Migration/Validation/MigrationValidationService.php
MalteJanz: But it would still clutter our log file right? And it's also extra processing work for something irrelevant (if it actually is ignored by the DAL which I'm expecting) 😅

larskemper: then let's get rid of it ➕ 😄

src/Migration/Validation/MigrationValidationService.php
larskemper: i removed the lookup. Also added distinction between optional & required field for field validation. And added an extra error log for missing/invalid translations that we should make fixable in the future (let the user create required translation in the admin)

src/Migration/Validation/MigrationValidationService.php
```diff
@@ -185,71 +224,84 @@ private function validateFields(MigrationValidationContext $validationContext):
                 $serializer = $field->getSerializer();
                 \iterator_to_array($serializer->encode($field, $entityExistence, $keyValue, $parameters), false);
```
jozsefdamokos: why is `iterator_to_array` needed here?

jozsefdamokos (approved):

src/Migration/Validation/MigrationValidationService.php
larskemper: the Serializers are returning `\Generator` ([ref](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/FieldSerializer/FieldSerializerInterface.php#L28)) that is not executed until actually iterated. So we need that to actually execute the encoding.

src/Migration/Validation/MigrationValidationService.php
```diff
@@ -22,31 +24,51 @@
...
-use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidFieldValueLog;
-use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidForeignKeyLog;
+use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidOptionalFieldValueLog;
+use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidRequiredFieldValueLog;
+use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidRequiredTranslation;
 use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationMissingRequiredFieldLog;
-use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationUnexpectedFieldLog;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\Service\ResetInterface;
+use function var_dump;
```
jozsefdamokos: I guess not needed

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -27,14 +29,16 @@
...
+use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidRequiredFieldValueLog;
+use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationInvalidRequiredTranslation;
 use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationMissingRequiredFieldLog;
-use SwagMigrationAssistant\Migration\Validation\Log\MigrationValidationUnexpectedFieldLog;
 use SwagMigrationAssistant\Migration\Validation\MigrationValidationResult;
 use SwagMigrationAssistant\Migration\Validation\MigrationValidationService;
 use SwagMigrationAssistant\Profile\Shopware54\Shopware54Profile;
 use SwagMigrationAssistant\Test\Mock\Gateway\Dummy\Local\DummyLocalGateway;
+use function array_map;
+use function var_dump;
```
jozsefdamokos: I guess not needed

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -166,10 +172,89 @@ public function testShouldValidateStructureAndFieldsValues(array $convertedData,
...
+        $baseData = [
+            'level' => 'error',
+            'code' => 'some_code',
+            'userFixable' => true,
+            'createdAt' => (new \DateTime())->format(\DATE_ATOM),
+        ];
+
+        yield 'missing id (null)' => [
+            $baseData,
+            MigrationValidationException::unexpectedNullValue('id')->getMessage(),
```
ennasus4sun, line 239:
```suggestion
            'convertedData' => $baseData,
            'expectedExceptionMessage' => MigrationValidationException::unexpectedNullValue('id')->getMessage(),
```

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -243,25 +366,13 @@ public static function entityStructureAndFieldProvider(): \Generator
...
-        ];
-
         yield 'fields - invalid type' => [
             [
                 ...$log,
                 'userFixable' => 'not_a_boolean',
             ],
             [
-                MigrationValidationInvalidFieldValueLog::class,
+                MigrationValidationInvalidOptionalFieldValueLog::class,
```
ennasus4sun, line 375: here the keys are missing, too

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -381,8 +481,299 @@ public static function associationProvider(): \Generator
...
+                ],
+            ],
+        ];
+
+        yield 'valid categories association (empty array)' => [
+            [
+                ...$baseProduct,
+                'categories' => [],
+            ],
+            [],
```
ennasus4sun, line 512:
```suggestion
            'convertedData' => [
                ...$baseProduct,
                'categories' => [],
            ],
            'expectedLogs' => [],
```

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
```diff
@@ -381,8 +481,299 @@ public static function associationProvider(): \Generator
...
+            'translations' => [
+                Defaults::LANGUAGE_SYSTEM => [
+                    'name' => 'Test Product',
+                ],
+            ],
+        ];
+
+        yield 'valid manufacturer association (null value)' => [
+            $baseProduct,
+            [],
```
ennasus4sun, line 617: same: keys missing

ennasus4sun (approved):

jozsefdamokos (approved):

tests/integration/Migration/Validation/MigrationValidationServiceTest.php
larskemper: when comes the day where I finally remember this 🤣. I need a phpstan rule for this

PR #107, opened 2025-12-19
refactor: select all performance

ennasus4sun:
Previously, clicking "Select All" fetched all log Id's from the database. For migrations with thousands of errors, this could lead to performance issues.

- before: "select-all" fetches all swag_migration_logging.id and stores them in `selectedLogIds` 
- now: "select-all" tracks a `selectAllMode` state and just visually selects all log-rows + disables checkbox ("all-or-nothing-principle") = no data is fetched

- fetching all unresolved entity ids in submission process: fetches and saves fixes in batches (limit is [Shopware default API limit value ](https://github.com/shopware/shopware/blob/d42b26a45c01bcdac9f1abdbd130d84c675dddee/src/Core/Framework/Resources/config/packages/shopware.yaml#L227)= 500)
- split submission logic into<br />`submitResolutionInBatches()` > `selectAllMode`<br/>and `submitResolutionForSelectedIds()` > unchanged submission flow

closes: https://github.com/shopware/shopware/issues/13672


src/Controller/HistoryController.php
```diff
@@ -164,12 +164,12 @@ public function getLogGroups(Request $request, Context $context): JsonResponse
...
     #[Route(
-        path: '/api/_action/migration/get-all-log-ids',
-        name: 'api.admin.migration.get-all-log-ids',
+        path: '/api/_action/migration/get-all-entity-ids',
+        name: 'api.admin.migration.get-all-entity-ids',
         methods: ['POST'],
         defaults: ['_acl' => ['swag_migration.viewer']]
     )]
-    public function getAllLogIds(Request $request): JsonResponse
+    public function getAllEntityIds(Request $request): JsonResponse
```
larskemper:
```suggestion
    public function getAllLogEntityIds(Request $request): JsonResponse
```

src/Controller/HistoryController.php
```diff
@@ -164,12 +164,12 @@ public function getLogGroups(Request $request, Context $context): JsonResponse
     }
 
     #[Route(
-        path: '/api/_action/migration/get-all-log-ids',
-        name: 'api.admin.migration.get-all-log-ids',
+        path: '/api/_action/migration/get-all-entity-ids',
+        name: 'api.admin.migration.get-all-entity-ids',
```
larskemper:
```suggestion
        name: 'api.admin.migration.get-all-log-entity-ids',
```

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -570,28 +570,30 @@ export default class MigrationApiService extends ApiService {
         );
     }
 
-    async getAllLogIds(
+    async getAllEntityIds(
```
larskemper:
```suggestion
    async getAllLogEntityIds(
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -223,6 +227,48 @@ export default Shopware.Component.wrapComponentConfig({
...
+
+                return;
+            }
+
+            const entities = entityIds.map((entityId) => this.createResolutionEntity(entityId));
+
+            await this.migrationFixRepository.saveAll(entities);
+        },
+
+        async submitResolutionInBatches() {
```
larskemper, line 250: lets avoid using `while (true)` here please 😅:

```suggestion
        async submitResolutionInBatches(iteration = 0): Promise<void> {
            const batchSize = 100;

            if (iteration >= 1000) {
                return;
            }

            const { entityIds } = await this.migrationApiService.getAllEntityIds(
                this.runId,
                this.selectedLog.code,
                this.selectedLog.entityName,
                this.selectedLog.fieldName,
                this.migrationStore.connectionId,
                batchSize,
                0,
            );

            if (entityIds.length === 0) {
                return;
            }

            const entities = entityIds.map((entityId: string) => this.createResolutionEntity(entityId));
            await this.migrationFixRepository.saveAll(entities);

            if (entityIds.length === batchSize) {
                await this.submitResolutionInBatches(iteration + 1);
            }
        },
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/swag-migration-error-resolution-modal.html.twig
```diff
@@ -99,14 +99,14 @@
                         <div v-if="resolvingComponent === 'DEFAULT'" class="swag-migration-error-resolution-modal__right-content-default">
                             <swag-migration-error-resolution-field
                                 :log="selectedLog"
-                                :disabled="loading || selectedLogIds.length === 0"
+                                :disabled="loading || (selectedLogIds.length === 0 && !selectAllMode)"
```
larskemper: use a dedicated computed property for this, that makes it simpler to read.

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -266,6 +312,7 @@ export default Shopware.Component.wrapComponentConfig({
 
         resetSelection() {
             this.selectedLogIds = [];
```
larskemper, line 323: Further notes:
- The “Deselect all” button in the data table isn’t working correctly: some logs remain selected, and the checkboxes are still disabled.
- The selected-items counter (e.g., “Edit: 100 Items”) behaves oddly in “Select all” mode, could it just use tableCount instead?

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
larskemper, line 460: lets early return here to reduce nesting and make it easier to read:

```suggestion
        async onSelectAllLogs() {
            if (!this.selectedLog || this.selectAllMode) {
                return;
            }

            this.applySelectionToGrid(true);

            await this.$nextTick();

            this.selectedLogIds = [];
            this.selectAllMode = true;
        },
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -473,6 +516,9 @@ export default Shopware.Component.wrapComponentConfig({
 
         onSelectionChanged(selection: Record<string, ResolutionModalRow>) {
             if (!selection || Object.keys(selection).length === 0) {
+                if (this.selectAllMode) {
```
larskemper: `if` is not really needed here if you set it to false anyway:
true => false
false => false

```suggestion
    this.selectAllMode = false;
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -491,6 +537,9 @@ export default Shopware.Component.wrapComponentConfig({
         },
 
         isRecordSelectable(item: ResolutionModalRow): boolean {
+            if (this.selectAllMode) {
+                return false;
+            }
             return !item.status;
```
larskemper:
```suggestion
            return !this.selectAllMode && !item.status;
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
larskemper:
```suggestion
        async onPageChange({ page, limit }: { page: number; limit: number }) {
            this.tablePage = page;
            this.tableLimit = limit;

            const wasSelectAllMode = this.selectAllMode;
            this.selectAllMode = false;

            await this.fetchLogs();
            await this.$nextTick();

            this.applySelectionToGrid(wasSelectAllMode);

            if (wasSelectAllMode) {
                await this.$nextTick();
                this.selectAllMode = true;
            }
        },
```

larskemper (dismissed):

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
ennasus4sun: mh, ... I dont know ... let me think about
`iteration >= 1000` is a safe-guard?
whats the problem with `while (true)`? for me the recursion is not feeling much better ...
whats with
```
while (hasMoreResults && iteration < 1000 ) {
// existing code
iteration++;
}
```
?

ennasus4sun: maybe a 

- request to a method which returns all unresolved log entity ids 
- and then calculate how much batches are needed
- and then looping with a foreach by using this as a limiter
would be the better way 🤔 

would just have one more request and query ...

src/Controller/HistoryController.php
```diff
@@ -196,17 +196,19 @@ public function getAllLogIds(Request $request): JsonResponse
         }
 
         $connectionId = $request->request->getAlnum('connectionId');
+        $limit = $request->request->getInt('limit', 100);
```
MalteJanz: Is this limit validated somewhere to have a upper bound and be greater than zero?
For example our admin HTTP API has a limit of `500`:
https://github.com/shopware/shopware/blob/d42b26a45c01bcdac9f1abdbd130d84c675dddee/src/Core/Framework/Resources/config/packages/shopware.yaml#L227

Why you could reuse from the service container like:
https://github.com/shopware/shopware/blob/2e918d106910b1510417510a74555f4b99d03e1d/src/Core/Framework/DependencyInjection/data-abstraction-layer.xml#L109

src/Controller/HistoryController.php
MalteJanz: Technically the `all` part is now also a bit misleading, because it only fetches one chunk up to the limit. Also it doesn't tell much about what conditions where applied to the ids. So I would suggest something more like:


```suggestion
    public function getEntityIdsOfLogWithoutFix(Request $request): JsonResponse
```

And change the other places + route path accordingly.

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
MalteJanz: I think the current solution with a simple while loop and exit condition is already quite nice and agree with @ennasus4sun that making it recursive doesn't seem much better.

But I also think that having a upper bound / safeguard is cleaner to not run the loop infinitly if we got a bug somewhere.

Just guessing that 1000 iterations is enough likely creates other problems in the future (what if customers have over 100k issues they want to batch solve and suddenly only the first 100k are solved).

I think the proper approach would be to fetch the actual total amount before starting this process (e.g. together with the get entity ids request), then use that as an upper bound / `for` loop. The good thing about that is it also makes it super easy to convert the loading spinner into a progress bar, which we likely want to have as I guess this process can take quite some time and if people get impatient starring at a loading spinner for minutes, they might think the page is broken and nothing will happen...

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -508,10 +557,23 @@ export default Shopware.Component.wrapComponentConfig({
             this.tablePage = page.page;
             this.tableLimit = page.limit;
 
-            await this.fetchLogs();
+            // temporarily disable select all mode to allow select checkboxes before disabling them
+            const wasSelectAllMode = this.selectAllMode;
+            if (wasSelectAllMode) {
+                this.selectAllMode = false;
+            }
```
MalteJanz:
```suggestion
            this.selectAllMode = false;
```

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
```diff
@@ -731,6 +749,76 @@ describe('module/swag-migration/component/swag-migration-error-resolution/swag-m
...
+
+            await wrapper.find('.swag-migration-error-resolution-modal__right-content-button').trigger('click');
+            await flushPromises();
+
+            const notifications = Object.values(Shopware.Store.get('notification').notifications);
+
+            expect(migrationApiServiceMock.getAllEntityIds).toHaveBeenCalledTimes(2);
+            expect(migrationFixRepositoryMock.saveAll).toHaveBeenCalledTimes(2);
+
+            expect(notifications).toHaveLength(1);
```
MalteJanz, line 829: This might get flaky. I would wait until the loading spinner disappeared before checking for the `toHaveBeenCalledTimes` and only afterwards fetch the notifications out of the store and validate it.

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
```diff
@@ -844,35 +960,128 @@ describe('module/swag-migration/component/swag-migration-error-resolution/swag-m
...
-            expect(migrationFixRepositoryMock.create).toHaveBeenCalled();
+            const notifications = Object.values(Shopware.Store.get('notification').notifications);
 
-            expect(migrationLoggingRepositoryMock.search).toHaveBeenCalledWith(
-                expect.objectContaining({
-                    page: 1,
-                    limit: 1,
-                    ids: ['log-id-3'],
-                }),
+            expect(notifications).toHaveLength(1);
```
MalteJanz, line 1116: see above, would also wait for the loading spinner to disappear here like a real user would instead of relying only on `await flushPromises();`

MalteJanz (changes requested):

src/Controller/HistoryController.php
MalteJanz: Additionally, how did you came up with the number `100` as the default and also actual used limit? I would suggest also experimenting with larger numbers up to `500` and see what the http request response times are. Also considering the payload of the requests / responses shouldn't be that huge. If they are still reasonable I can imagine going with a larger number makes the process faster, e.g. if you can process 50k entities (100 * 2 requests vs 500 * 2 requests) in X vs Y amount of time

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
ennasus4sun: - regarding "deselect all" issue: I cant reproduce this on my local instance (neither with admin-watcher enabled nor without); maybe we can have a look together later
- regarding "selected-items counter": yes, this is an old issue. I will try your suggestion but will create a new issue for that if it is not working

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
ennasus4sun: true, of course 👍🏻

src/Controller/HistoryController.php
ennasus4sun: for now I use the Shopware default limit without making some timing tests.
100 was just a random number ...

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
ennasus4sun: to be honest, Iam not sure how `flushPromises()` works when async calls created in a loop 🙄.
any ideas? see my current "fix" (cc @larskemper )

src/Controller/DataProviderController.php
```diff
@@ -54,7 +55,7 @@ public function __construct(
     #[Route(
         path: '/api/_action/data-provider/get-environment',
         name: 'api.admin.data-provider.get-environment',
-        defaults: ['_acl' => ['admin']],
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['admin']],
```
DennisGarding, line 58: Like that changes! 👍

DennisGarding (approved):

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
larskemper: can't reproduce the deselect thing anymore, so i guess it was something local 😅

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
larskemper: Missed that, we should just simply flush the promises here.
`flushPromises()` just flushes every pending promise that exists, so the loop should not make a difference.

ennasus4sun: yes, I think the important part is "that exists".
I will remove the `while(wrapper.vm.submitLoading){flushPromises()}` part for now and just expect the loader is gone

larskemper: what I meant is that just a single `await flushPromises()` is enough. It already resolves all created promises... then you are sure that the loader is gone

ennasus4sun: changed my mind 😅 fails without this part

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
ennasus4sun: regarding the "selected items counter": tested with `tableTotal` (`{ count: selectAllMode ? tableTotal : selectedLogIds?.length }`) but its not working. Will create a ticket for that.

larskemper: You don't need to. I will just do it with: https://github.com/shopware/shopware/issues/14490

ennasus4sun: https://github.com/shopware/shopware/issues/14515

src/Controller/HistoryController.php
```diff
@@ -197,16 +267,17 @@ public function getAllLogIds(Request $request): JsonResponse
...
         $connectionId = $request->request->getAlnum('connectionId');
 
-        $logIds = $this->logGroupingService->getAllLogIdsByCodeAndEntity(
+        $logEntityIds = $this->logGroupingService->getLogEntityIdsWithoutFixByCodeAndEntity(
             $runId,
             $code,
             $entityName,
             $fieldName,
-            !empty($connectionId) ? $connectionId : null
+            $request->request->getInt('limit', $this->maxLimit),
```
MalteJanz: There is still no validation on this query parameter at all. Where is it validated that it's greater than zero and less than or equals to `$this->maxLimit`?

My previous comment on this: https://github.com/shopware/SwagMigrationAssistant/pull/107#discussion_r2713029197

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -570,22 +570,55 @@ export default class MigrationApiService extends ApiService {
...
+                return ApiService.handleResponse(response);
+            });
+    }
+
+    async getLogEntityIdsWithoutFix(
+        runId: string,
+        code: string,
+        entityName: string,
+        fieldName: string,
+        connectionId?: string,
```
MalteJanz, line 579: no `limit` query parameter here? I'm fine with leaving it as `null` / `undefined` and letting the server decide but I think this API service should mirror the PHP endpoint and show all possible parameters

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -223,6 +231,52 @@ export default Shopware.Component.wrapComponentConfig({
...
+        async submitResolutionInBatches() {
+            const { count, limit } = await this.migrationApiService.getUnresolvedLogsBatchInformation(
+                this.runId,
+                this.selectedLog.code,
+                this.selectedLog.entityName,
+                this.selectedLog.fieldName,
+                this.migrationStore.connectionId,
+            );
+
+            const iterations = Math.ceil(count / limit);
```
MalteJanz, line 259: In this case thought, you are assuming a certain limit is used in the backend and should also give that to the HTTP call below to ensure its the same with your assumption

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -223,6 +231,52 @@ export default Shopware.Component.wrapComponentConfig({
...
+                    this.selectedLog.fieldName,
+                    this.migrationStore.connectionId,
+                );
+
+                const entities = entityIds.map((entityId: string) => this.createResolutionEntity(entityId));
+
+                // each batch must be completed before fetching the next
+                // eslint-disable-next-line no-await-in-loop
+                await this.migrationFixRepository.saveAll(entities);
+            }
```
MalteJanz, line 278: Nitpick, I liked the `while` loop and It's also possible to do it without knowing any `limit` (you are relying on the server for that anyways) and with proper early break if the response is empty already for whatever reason, e.g. something like this
```TS
let offset = 0;

while (offset < count) {
    // each batch must be completed before fetching the next
    // eslint-disable-next-line no-await-in-loop
    const { entityIds, total } = await this.migrationApiService.getLogEntityIdsWithoutFix(
        this.runId,
        this.selectedLog.code,
        this.selectedLog.entityName,
        this.selectedLog.fieldName,
        this.migrationStore.connectionId,
        // for the limit, maybe just let the server decide on a default?
    );

    if (entityIds.length === 0) break;

    const entities = entityIds.map((entityId: string) => this.createResolutionEntity(entityId));

    // each batch must be completed before fetching the next
    // eslint-disable-next-line no-await-in-loop
    await this.migrationFixRepository.saveAll(entities);

    offset += entityIds.length;
}
```

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
```diff
@@ -731,6 +757,94 @@ describe('module/swag-migration/component/swag-migration-error-resolution/swag-m
...
+
+            // wait until all promisses in the loop are resolved
+            // 3 batches with 2 promises each = max 6 iterations + some more to be safe
+            let iterations = 10;
+            while (wrapper.vm.submitLoading && iterations > 0) {
+                // async calls created in a loop; we need to wait for them to be resolved
+                // eslint-disable-next-line no-await-in-loop
+                await flushPromises();
+                iterations -= 1;
+            }
```
MalteJanz: I don't like this, and I'm wondering if this is needed at all or if we can find a cleaner solution.

Regarding my previous two [comments](https://github.com/shopware/SwagMigrationAssistant/pull/107/changes#r2713234757) on this topic, I'm sorry but my suggestion was wrong and I though about playwright, it's not easily possible to wait for an element to disappear in Jest in a clean way.

I guessed a single call to `flushPromises` should be fine but now, I'm not sure if it might only be flaky 🤔 

sidenote: `flushPromises` is defined as follows:
https://github.com/vuejs/test-utils/blob/d7e8a3b35b3e4a8079a9c44d15a6cd7372b1dc37/src/utils/flushPromises.ts#L1-L8

And we likely don't recommend not awaiting in a loop for a reason 😅.

So I looked a bit deeper into this with a little experiment:

https://jsfiddle.net/o9087wty/3/

But I guess you have to do it with a real Jest and vue test environment, because in that tests `flushPromises` didn't wait for anything 🤔 

TLDR: what happens if you just call `await flushPromises` here once?

MalteJanz (changes requested):

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
ennasus4sun: hm, yes. But I think I will leave it as it is at the moment. It's also easier to have control in the jest-test about that value.

src/Controller/HistoryController.php
ennasus4sun: ok, I added both checks.
one question: shouldn't the user decide for themselves if they want to use a higher limit than our default value?

tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
ennasus4sun: tested that 10 times locally: all are successful ✅ 
\+ this pipeline ✅ <br />
I still have the assertion inside that the loading-spinner is gone before checking the notification message.
We can try like it is.

MalteJanz (approved):

src/Controller/HistoryController.php
MalteJanz: > one question: shouldn't the user decide for themselves if they want to use a higher limit than our default value?

In this case yes it's true that the user can't use a higher value than our default, because we already use `maxLimit` as our default... 
But technically he could still adjust the `shopware.yaml` and increase that `maxLimit` for the overall admin API if he really wants to.

In general you want to have an upper bound to make sure nobody sends a huge limit and tries to overload the server with many of these requests. And also because technically there is a limit of processing time (e.g. the request will terminate after 30s) as a last resort, but it's much better to contain the processing time to be something reasonable by properly validating the request parameters to stay in reasonable limits and reporting a proper error otherwise.

vintagesucks (approved):

PR #112, opened 2026-01-12
fix: use item-specific tax in order conversion

vintagesucks:
uses each line item's own `tax_rate` instead of the order-level tax rules


src/Profile/Shopware/Converter/OrderConverter.php
```diff
@@ -875,6 +875,8 @@ protected function getShippingMethod(string $shippingMethodId): ?string
     }
 
     /**
+     * @deprecated tag:v16.0.0 - parameter $taxRules will be removed
```
vintagesucks, line 878: not sure how to properly do this without phpstan throwing a tantrum, hence the ignored errors 😅

vintagesucks:
#114

PR #113, opened 2026-01-12
feat: Compatible with product bundle

### LINKED ISSUE #13167: Compatible with migration assistant
### Technical TODO

Bundle product (product with type = grouped_bundle) should be excluded when migrating between different SW instances
`States` field should be migrated to `type` field 

Note: Remember to check if the field type exists or the column type exists in the database

tamvt:
In this PR:

- We need to migrate product states <-> product type and product type <-> product type
- Product type PR: https://github.com/shopware/shopware/pull/13554/files
- We need to exclude products with type grouped_bundle when migrating because the product bundles are not yet compatible with MIG yet.


nguyenytran (approved):

vienthuong (approved):

src/Profile/Shopware6/Converter/ProductConverter.php
```diff
@@ -58,6 +74,16 @@ protected function convertData(array $data): ConvertStruct
...
 
+        // Check if product already exists (was previously migrated) before creating mapping
+        // This is needed because the 'type' field has Immutable flag and can only be set on create
+        $existingMapping = $this->mappingService->getMapping(
+            $this->connectionId,
+            DefaultEntities::PRODUCT,
+            $data['id'],
+            $this->context
+        );
+        $isUpdate = $existingMapping !== null;
```
MalteJanz: At first look, this is a clever attempt to determine if this entity will run into an `insert` or `update`. But unfortunately it contains a flaw 🙈 .

Yes on the first migration run this will tell you if the product "was previously attempted to migrate" and thus give the the wanted result.

But if the migration for this entity somehow fails:
- either further down in this converter
- in the validation layer that comes after - which we currently implement [in our feature branch](https://github.com/shopware/SwagMigrationAssistant/pull/41)
- or later while writing the entity to the corresponding SW6 table

it would mean the next migration run (retry) would still consider this entity as an "update", because the mapping already exists, which produces the wrong outcome in this case.

We have to think further about this, to not introduce yet another [N+1 Query performance issue](https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it) 😬

src/Profile/Shopware6/Converter/CustomFieldSetConverter.php
```diff
@@ -27,12 +27,43 @@ protected function convertData(array $data): ConvertStruct
     {
         $converted = $data;
 
+        // Check if custom field set already exists (was previously migrated) before creating mapping
+        // This is needed because the 'name' field has Immutable flag and can only be set on create
```
MalteJanz: Sorry for not raising this concern earlier:

Do you really think it's a good idea to have such a [ImmutableFlag](https://github.com/shopware/shopware/blob/6331a8d4adc5b2832ac9f671e33737e463e8b4e7/src/Core/Framework/DataAbstractionLayer/Field/Flag/Immutable.php#L10) ? (Was introduced [in this PR](https://github.com/shopware/shopware/pull/13957) )

Because I already see the extra work it creates for us here, every entity that uses that flag needs an extra special treatment in the migration and if someone adds that flag to other fields in the future (or introduces such fields), we will run into issues again.

It basically makes "upserts" (insert / update) really difficult to do for such entities, because you want to create them if they don't exists yet, or update them otherwise, so you need the field to be defined but also not defined at the same time... 🤯 

I also can't imagine that it's nice to work with such an API, e.g. imagine:
- You read the data of an existing entity over our admin API
- You modify some value
- You want to update that entity with the same payload

I think this was possible in the past (correct me if I'm wrong), but now you need to remove certain fields from your payload to send the update request, which is really annoying.

Of course in our administration codebase this is a non-issue, because there we use some extra logic / magic to build a changeset diff and only send that back to the server, but do you really expect every other API client to do the same or treat it differently like we do here in the migration? 😬

I would suggest either:

1. Allowing updates in the DAL when the value of the `Immutable` field stays the same (is unchanged), as it should be in these cases like here in the "upsert" case for migration. Otherwise it would be a valid error if the migration tries to change such a value
2. Or "silently" ignore that field in the DAL update case, without throwing an error. Maybe it should still create a log entry though 🤷‍♂️ 

I'm open for other suggestions as well, but I think we should have a better answer for the "upsert" case than just deal with it by throwing extra code at it. The Migration-Assistant makes heave use of "upsert" and basically writes all it's data that way:
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/Migration/Writer/AbstractWriter.php#L60

cc: @patzick @keulinho what's your opinion on this?

MalteJanz (changes requested):
Thanks for reaching out and taking care of this change 👍 . But I think we need to discuss a little further about the chosen approach and figure out a more robust one 🙂

src/Profile/Shopware6/Converter/CustomFieldSetConverter.php
vienthuong: That do makes sense, I was also thinking about the update should allow the payload with immutable key if the value doesn't change, but at the query time, we haven't aware of the current value in the database, fetching it before wouldn't be ideal, I will have a look further at this problem and your suggestion and provide an improvement for the flag. 

For an external API consumer, I think ideally they don't want to send the whole entities; a subset of what changes is more practical and lightweight. But there might be cases where they want to resend the entire payload and catch the immutable error, as in your mentioned scenario.

Another simple option is to bypass the immutable check for system context scope; it will work for examples like MIG, but not if they update the entities via the admin api.

vienthuong: > Allowing updates in the DAL when the value of the Immutable field stays the same (is unchanged), as it should be in these cases like here in the "upsert" case for migration. Otherwise it would be a valid error if the migration tries to change such a value

This should be possible, I added a draft PR here: https://github.com/shopware/shopware/pull/14422

src/Profile/Shopware6/Converter/ProductConverter.php
```diff
@@ -235,4 +253,40 @@ private function checkDefaultCurrency(array &$source, string $key): void
             $source[$key][] = $defaultPrice;
         }
     }
+
+    /**
+     * @param array<string, mixed> $converted
+     */
+    private function convertStatesToType(array &$converted): void
+    {
+        if (!$this->hasTypeColumn()) {
```
MalteJanz: Not sure if this is actually needed (I would need to test this myself). But what happens if you provide an extra field in a `repository-upsert(...)`, that doesn't exist in the definition / DB? I guess it might get ignored.

If that's the case we don't even have to check for the column to exists and could just always add / set that `type` to an appropriate value

tamvt: Good catch 👍 .
I checked again and found that it had been skipped, so I removed the check.

src/DataProvider/Provider/Data/ProductProvider.php
```diff
@@ -113,6 +122,34 @@ public function getProvidedData(int $limit, int $offset, Context $context): arra
...
+    }
+
+    private function hasTypeColumn(): bool
+    {
+        if ($this->hasTypeColumn !== null) {
+            return $this->hasTypeColumn;
+        }
+
+        $columns = $this->connection->createSchemaManager()->listTableColumns('product');
+        $this->hasTypeColumn = isset($columns['type']);
```
MalteJanz: Sorry for not mentioning this earlier, but I think these calls are quite expensive performance wise (and I know you already memoize / cache them, but the provider is usually only called once per request, so this doesn't provide many benefits).

Because this product provider is querying data via the DAL and not plain SQL, I think we could also rely on the DAL definition here to check if the column exists (and can be used in the Criteria). e.g. something like

```suggestion
        $this->hasTypeColumn = isset($this->productRepo->getDefinition()->getField('type'));
```

which means you can get rid of the `Connection` dependency and this is likely quite a bit faster because it doesn't have to reach out to the database to query all the fields.

src/DataProvider/Provider/Data/ProductProvider.php
```diff
@@ -20,12 +23,17 @@
 #[Package('fundamentals@after-sales')]
 class ProductProvider extends AbstractProvider
 {
+    private const BUNDLE_PRODUCT_TYPE = 'grouped_bundle';
+
+    private ?bool $hasTypeColumn = null;
```
MalteJanz: As mentioned above, the provider is usually only called once for a single request, so this memoize / cache doesn't provide many benefits => I would remove it.

If you would want to keep it, you would also need to implement the `ResetInterface` to ensure the value doesn't stick around too long on certain environments / deployments.

larskemper (approved):

vienthuong (approved):

MalteJanz (approved):

PR #114, opened 2026-01-12
fix: use item-specific tax in order conversion

vintagesucks:
https://github.com/shopware/shopware/issues/13898

uses each line item's own `tax_rate` instead of the order-level tax rules

targeting `feature/migration-logging-refactor` (next major) to avoid the unnecessarily complex deprecation path


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 16.0.0
 
+- [BREAKING] Removed parameter `taxRules` from method `getLineItems` of `SwagMigrationAssistant\Profile\Shopware\Converter\OrderConverter`
```
MalteJanz: Sorry, I should have explained this as well while we were talking about this change:

In our plugins the `CHANGELOG.md` / `CHANGELOG_de-DE.md` are user facing and will end up in the changelog of our plugin store: https://store.shopware.com/en/swag257162657297f/migration-assistant.html

For technical (breaking) changes we introduced this file [UPGRADE.md](https://github.com/shopware/SwagMigrationAssistant/blob/feature/migration-logging-refactor/UPGRADE.md) (which also is only maintained in English to keep it simple).

But in the user facing changelog you could still mention that this bug was fixed and different tax rates on line items of orders are now properly migrated from SW5 🎉 

*It's a completely different process than the one in the core shopware Repo. I don't know if we documented it somewhere, but I hope this gets an overhaul at one point as well 🙈

vintagesucks: No problem, something like this? [`aef1cd8` (#114)](https://github.com/shopware/SwagMigrationAssistant/pull/114/commits/aef1cd8ce5e1943bce63fea9cccc56b7ac133593)

src/Profile/Shopware/Converter/OrderConverter.php
vintagesucks, line 1: TODO on merge: Add `resolves shopware/shopware#13898` to `#41`

CHANGELOG.md
vintagesucks: Oh, did not see your edit in time, and this? 🙈  

[`003153a` (#114)](https://github.com/shopware/SwagMigrationAssistant/pull/114/commits/003153ac8d53c043c6a69085685c692763e36869)

MalteJanz (approved):

larskemper (approved):

PR #117, opened 2026-01-21
refactor: migration validation

larskemper:
Overall refactor of the migration validation behavior. Should lead to much less needed fixes and more accurate / actually fixable errors.

sorry, for the big pr, but refactor was needed 😅. Reach out if I can explain anything and please give me your ideas for better validation!

### Changes (file by file)

- `src/Controller/DataProviderController.php`, `src/Controller/HistoryController.php`, `src/Controller/PremappingController.php`, `src/Controller/StatusController.php`:
   Introduced all common route definition constants to unify with platform

- [Added] `src/Controller/ErrorResolutionController.php`:
   Added two new routes:
   - `/error-resolution/validate`: provides validation for user created fixes to make sure the fix matches original migration validation
   - `/error-resolution/example-field-structure`: in case we can not determine the field type in the administration this route provides an example field structure of json association fields, so the user can follow this structure to create a valid fix.

- `src/Migration/ErrorResolution/MigrationErrorResolutionService.php`: added missing early return if no fixes could be loaded to avoid event spam and unnecessary code execution

- [Added] `src/Migration/ErrorResolution/MigrationFieldExampleGenerator.php`: the generator used for the example field structure route. It takes in a field and tries to build a accurate example structure for the user

- [!!!] `src/Migration/ErrorResolution/MigrationFix.php`: see the comment

- `src/Migration/Mapping/MappingService.php`: removed the mapping lookup. We decided to not validate of a given association is actually existing and just validate the given data

- `src/Migration/Service/MediaFileProcessorService.php`: add where clause to only fetch unprocessed media

- `src/Migration/Service/MigrationDataWriter.php`: change to `->count()` to return number of entities that are processed in the current batch, not the absolute total

- [!!!] `src/Migration/Validation/MigrationEntityValidationService.php`:
   - Deleted association mapping lookup
   - New definition of required field: not system managed (createdAt, ...), actually required in db, no default value
   - Validated nested converted data
   - More naive way of validating association field, cause the before validation by serializer needed context that we actually can't provide in the migration. Now more like a structural validation of association data.
   - Move Exceptions to validation domain for better separation
 
- [!!!] `src/Migration/Validation/MigrationFieldValidationService.php`:
   - same as entity validator: more naive validation, support of nested fields
   - splited from entity validator to provide standalone field validation for the error resolution controller 

- `src/Migration/Validation/MigrationValidationResult.php`: added log key generation to avoid multiple log for the same invalidation (avoid multiple fixes for same issue)

- `src/Profile/Shopware/Converter/OrderConverter.php`: bug fix (custom field was set to `null` before, instead of just not adding it to the converted data)

- `src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-field/swag-migration-error-resolution-field-scalar/index.ts`: added display of error messages that are provides by the validation route

- `src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-field/swag-migration-error-resolution-field/index.ts`: added auto fetch of example field structure in case the admin could not determine a specific field type

- `src/Resources/app/administration/src/module/swag-migration/service/swag-migration-error-resolution.service.ts`: support of nested field paths and overall simplification


src/Migration/ErrorResolution/MigrationFix.php
```diff
@@ -50,38 +50,49 @@ public static function fromDatabaseQuery(array $data): self
...
+
+        // last segment of the path, "normal" set operation
+        if (empty($remainingPath)) {
+            $data[$key] = $value;
+
+            return;
+        }
+
+        // key points to a list, apply to all items in the list
+        if (isset($data[$key]) && \is_array($data[$key]) && \array_is_list($data[$key])) {
```
larskemper: example case for this:
entity `number_range` with converted data like this:

```php
[
    'id' => 'uuid'
    // ...
    'numberRangeSalesChannels' => [
        [
            // ...
             'typeId' => null // (invalid, has to be set)
        ],
        [
            'typeId' => null
        ]
    ]
]
```

with this solution the fix would be applied for all sub child of `numberRangeSalesChannels`.
To support individual fixes for sub childs we would have to rebuild quite a lot of the error resolution to a path syntax like `parent.sub.[index].property` which we currently are not supporting and would cause problems in the ui (because we group by fieldname).

What do you think? Leave it like this or maybe not support it at all for the moment (would mean these entities can not be migrated, like before anyway)

tests/integration/Migration/Validation/MigrationEntityValidationServiceTest.php
```diff
@@ -0,0 +1,890 @@
...
+    {
+        $baseData = [
+            'level' => 'error',
+            'code' => 'some_code',
+            'userFixable' => true,
+            'createdAt' => (new \DateTime())->format(\DATE_ATOM),
+        ];
+
+        yield 'missing id (null)' => [
+            $baseData,
```
larskemper: I will add named index 😅

src/Controller/DataProviderController.php
```diff
@@ -32,7 +34,7 @@
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\Routing\Attribute\Route;
 
-#[Route(defaults: ['_routeScope' => ['api']])]
+#[Route(defaults: [PlatformRequest::ATTRIBUTE_ROUTE_SCOPE => [ApiRouteScope::ID]])]
```
DennisGarding, line 37: Like that changes 👍

src/Migration/ErrorResolution/MigrationErrorResolutionService.php
```diff
@@ -30,6 +31,8 @@ public function __construct(
 
     /**
      * @param array<int|string, array<int|string, mixed>> $data
+     *
+     * @throws Exception
```
DennisGarding: `@throws Exception` can be removed. In the function is no throw

src/Migration/ErrorResolution/MigrationErrorResolutionService.php
```diff
@@ -76,6 +83,8 @@ public function applyFixes(array &$data, string $connectionId, string $runId, Co
 
     /**
      * Loads fixes from the database and populates them in the context.
+     *
+     * @throws Exception
```
DennisGarding: Same here

DennisGarding (approved):

src/Migration/Validation/Exception/MigrationValidationException.php
```diff
@@ -0,0 +1,106 @@
...
+            ['fieldName' => $fieldName]
+        );
+    }
+
+    public static function invalidId(string $entityId, string $entityName): self
+    {
+        return new self(
+            Response::HTTP_INTERNAL_SERVER_ERROR,
+            self::VALIDATION_INVALID_ID,
+            'The id "{{ entityId }}" for entity "{{ entityName }}" is not a valid Uuid',
```
vintagesucks:
```suggestion
            'The id "{{ entityId }}" for entity "{{ entityName }}" is not a valid UUID',
```

src/Migration/ErrorResolution/MigrationFix.php
```diff
@@ -50,38 +51,49 @@ public static function fromDatabaseQuery(array $data): self
      */
     public function apply(array &$item): void
     {
-        /*
```
jozsefdamokos, line 53: We should keep this I think

src/Migration/Validation/Log/MigrationValidationRequiredFieldValueInvalidLog.php
```diff
@@ -20,11 +20,11 @@ public function isUserFixable(): bool
 
     public function getLevel(): string
     {
-        return self::LOG_LEVEL_WARNING;
+        return self::LOG_LEVEL_ERROR;
     }
 
     public function getCode(): string
     {
-        return 'SWAG_MIGRATION_VALIDATION_UNEXPECTED_FIELD';
```
vintagesucks, line 28: Snippets are still present

src/Migration/ErrorResolution/MigrationFix.php
```diff
@@ -50,38 +51,49 @@ public static function fromDatabaseQuery(array $data): self
...
+    }
+
+    /**
+     * Recursively applies the fix value to the specified path.
+     * When encountering a list (numerically-indexed array), applies the fix to all items.
+     *
+     * @param array<string|int, mixed> $data
+     * @param array<int, string> $remainingPath
+     */
+    private function applyToPath(array &$data, array $remainingPath, mixed $value): void
```
jozsefdamokos: Maybe better to call the function param simply `$path` and set `$remainingPath` inside the function

src/Migration/Validation/MigrationEntityValidationService.php
```diff
@@ -0,0 +1,466 @@
...
+use Symfony\Contracts\Service\ResetInterface;
+
+/**
+ * @internal
+ */
+#[Package('fundamentals@after-sales')]
+class MigrationEntityValidationService implements ResetInterface
+{
+    /**
+     * System managed that managed by Shopware and should not be validated as required fields
```
vintagesucks:
```suggestion
     * System managed fields are managed by Shopware and should not be validated as required fields
```

or

```suggestion
     * System managed fields that are managed by Shopware and should not be validated as required fields
```

src/Migration/ErrorResolution/MigrationFix.php
jozsefdamokos: Is this a realistic scenario with the number ranges?

Im fine with this solution but we should make it clear that this will happen to the user with a warning or tooltip

src/Migration/Validation/MigrationEntityValidationService.php
```diff
@@ -0,0 +1,466 @@
...
+
+        $this->loggingService->saveLogging($validationContext->getContext());
+
+        return $validationContext->getValidationResult();
+    }
+
+    /**
+     * Validates that all required fields are present in the converted data.
+     *
+     * @throws \Exception|Exception
```
jozsefdamokos: Is this needed? Same question for all methods in this class 😉

src/Migration/Validation/MigrationEntityValidationService.php
```diff
@@ -0,0 +1,466 @@
...
+    public function reset(): void
+    {
+        $this->requiredDefinitionFieldsCache = [];
+    }
+
+    /**
+     * @param array<string, mixed>|null $convertedEntity
+     * @param array<string, mixed> $sourceData
+     *
+     * @throws \Exception
```
jozsefdamokos: Is this needed?

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-field/swag-migration-error-resolution-field-unhandled/index.ts
```diff
@@ -27,48 +26,40 @@ export default Shopware.Component.wrapComponentConfig({
...
-            }
-
-            try {
-                const value = JSON.parse(this.fieldValue);
-                this.error = null;
-
-                return value;
-            } catch {
-                this.error = {
-                    detail: this.$tc('swag-migration.index.error-resolution.errors.invalidJsonInput'),
```
vintagesucks, line 67: Snippets are still present

src/Migration/Validation/MigrationFieldValidationService.php
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        $this->validateScalarField($field, $value, $isRequired, $entityDefinition, $context);
+    }
+
+    /**
+     * Resolves a potentially nested field path (e.g., "prices.shippingMethodId") to its target.
+     * Traverses through association fields to find the final entity definition and field.
+     *
+     * @throws \Exception
```
jozsefdamokos: Is this needed? Same question for the other methods where `@throws` is used

vintagesucks (approved):
🚀🚀🚀

src/Migration/Validation/MigrationEntityValidationService.php
```diff
@@ -0,0 +1,466 @@
...
+            return;
+        }
+
+        if ($field instanceof OneToManyAssociationField || $field instanceof ManyToManyAssociationField) {
+            $referenceDefinition = $field instanceof ManyToManyAssociationField
+                ? $field->getToManyReferenceDefinition()
+                : $field->getReferenceDefinition();
+
+            foreach ($value as $nestedEntityData) {
+                $this->validateNestedEntityData(
```
jozsefdamokos: Can we not use the same method `validateFieldValues` here? From the docblocks it seems to me that these two methods do a similar job.

tests/integration/Migration/Validation/MigrationFieldValidationServiceTest.php
jozsefdamokos, line 1: Some tests are missing for validating association structures

jozsefdamokos (changes requested):

jozsefdamokos:
Looks good! Great job!

src/Migration/ErrorResolution/MigrationErrorResolutionService.php
larskemper: `fetchAllAssociative()` inside of `loadFixes()` can throw

src/Migration/ErrorResolution/MigrationFix.php
larskemper: it just feels not completely right, but it seemed quite common. The problem is that if we don't to this adjustment we ask the user to provide a fix that we actually are not supporting to fix. So we would need new extra logic to prevent them to appear "fixable"

src/Migration/Validation/MigrationEntityValidationService.php
larskemper: i mean technically definitely not, but documenting the most ide's will highlight the usage of the method "this can throw..." and that's not unrelevant for the validation, cause all thrown errors should be catched to create logs / fixes. But I can also live with removing them 😄

vintagesucks: ~~For me, this line resulted in a "Why both `\Exception` and `Exception`?" question 😄~~

nevermind 😅 
> `use Doctrine\DBAL\Exception;`

src/Migration/ErrorResolution/MigrationErrorResolutionService.php
DennisGarding: yes, but normaly annotate exceptions if its realy thrown in the mehtod. So fetchAllAssociative need the `@throw` tag

src/Migration/Validation/MigrationEntityValidationService.php
larskemper: `validateFieldValues()` is used for the root entity validation and `validateNestedEntityData()` for the nested entity. I think the duplicated code is acceptable, but I will rename them to more obvious

src/Migration/Validation/MigrationEntityValidationService.php
jozsefdamokos: But there is no `throw` in these methods. The exceptions are caught inside them, they are not re-thrown. And if we keep the line @vintagesucks mentioned maybe we can add an alias `DbalException`

src/Migration/Validation/MigrationEntityValidationService.php
jozsefdamokos: I mean ideally you would somehow make the same method work for both entities, whether its root or nested. But I can see that this can be quite difficult. So Im fine with it like this.

jozsefdamokos (approved):

src/Migration/ErrorResolution/MigrationFieldExampleGenerator.php
```diff
@@ -0,0 +1,211 @@
...
+                'price' => 0.1,
+                'quantity' => 1,
+                'isCalculated' => false,
+                'taxRules' => [
+                    [
+                        'taxRate' => 0.1,
+                        'percentage' => 0.1,
+                    ],
+                ],
+            ],
```
MalteJanz, line 159: Nitpick: how can we make sure that future core changes in these schema's are also reflected here? e.g. in this case changes to this serializer 🤔
https://github.com/shopware/shopware/blob/8378f23333b6ca504cbe8d95846cd7364b4e9caa/src/Core/Framework/DataAbstractionLayer/FieldSerializer/PriceDefinitionFieldSerializer.php

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -73,6 +73,7 @@ public function writeData(MigrationContextInterface $migrationContext, Context $
         $criteria->addFilter(new EqualsFilter('entity', $dataSet::getEntity()));
         $criteria->addFilter(new EqualsFilter('runId', $migrationContext->getRunUuid()));
         $criteria->addFilter(new EqualsFilter('convertFailure', false));
+        $criteria->addFilter(new EqualsFilter('written', false));
         $criteria->setOffset($migrationContext->getOffset());
```
MalteJanz: This likely break the pagination logic for the writing step, e.g. the offset is not stable if you filter by `written=false`. Or to ask differently, what is the intention behind this change?

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -132,7 +133,7 @@ public function writeData(MigrationContextInterface $migrationContext, Context $
             }
             unset($data);
 
-            return $migrationData->getTotal();
+            return $migrationData->count();
```
MalteJanz: Same here, why this change while `getTotal` worked perfectly fine? It should always count up the offset by the amount of entities processed, regardless if they were successfull or not, to have the progress bar move properly and align with it's precomputed overall total

src/Migration/Service/MigrationDataWriter.php
```diff
@@ -163,7 +164,7 @@ public function writeData(MigrationContextInterface $migrationContext, Context $
             $context
         );
 
-        return $migrationData->getTotal();
+        return $migrationData->count();
```
MalteJanz: see above

src/Migration/Validation/MigrationFieldValidationService.php
```diff
@@ -0,0 +1,308 @@
...
+        $field = clone $field;
+
+        if ($field instanceof AssociationField) {
+            $this->validateAssociationStructure($field, $value, $entityDefinition->getEntityName());
+
+            return;
+        }
+
+        $this->validateScalarField($field, $value, $isRequired, $entityDefinition, $context);
+    }
```
MalteJanz, line 81: Nitpick: This looks expensive, e.g. why does the field need to be "resolved" for every value we want to validate? E.g. you might have a batch of 100 products and all have the same fields which need to be validated. We should do some performance profiling in our feature branch after this is merged and maybe apply some optimizations here

src/Migration/Validation/MigrationFieldValidationService.php
```diff
@@ -0,0 +1,308 @@
...
+        bool $isRequired = true,
+    ): void {
+        if (!$this->definitionRegistry->has($entityName)) {
+            return;
+        }
+
+        $resolved = $this->resolveFieldPath($entityName, $fieldName);
+
+        if ($resolved === null) {
+            return;
```
MalteJanz, line 66: Nitpick: Also "silent" early returns? Would be nice to add a comment in these cases that make it clear it's fine, e.g.

```suggestion
        if ($resolved === null) {
            // field does not exists in this SW6 installation, so it will be skipped by the DAL => it's fine to ignore here
            return;
```

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
```diff
@@ -231,6 +226,49 @@ export default Shopware.Component.wrapComponentConfig({
...
+            if (validationError) {
+                this.createNotificationError({
+                    message: this.$tc(`swag-migration.index.error-resolution.errors.${validationError}`),
+                });
+
+                return false;
+            }
+
+            const serializationError = await this.migrationApiService
+                .validateResolution(this.selectedLog.entityName, this.selectedLog.fieldName, this.fieldValue)
```
MalteJanz, line 245: Nitpick: what's the difference between `validateFieldValue` and `validateResolution` Both calls look almost identical to me and have the same parameters. Can't this be done in a single backend call?

MalteJanz (approved):
@larskemper appreciate the effort you put into this and we are moving in the right direction 💪 

I only had a rough look because of the size of the PR and left some comments, but overall I'm fine with merging this into our feature branch (feel free to create follow up issues / PRs for my comments or fix them before merging, your decision. The nitpicks are just nice to have / some further discussion points).

But I don't think we reached a good state for merging the feature branch into `trunk` yet. I noticed there are still quite a few confusing errors being reported when I migrate from my SW5 shop with demo data, compared to the `trunk` state. Most likely we:
- might broke some converters logic that is producing validation errors now
- the validation logic still reports false positives

which we should investigate further before release 🙂

src/Migration/ErrorResolution/MigrationFieldExampleGenerator.php
larskemper: we could write a test that uses the serializer the way we expected it here and check if it's failing

src/Migration/Validation/MigrationFieldValidationService.php
larskemper: https://github.com/orgs/shopware/projects/43/views/2?pane=issue&itemId=3897161358&issue=shopware%7Cshopware%7C14752

src/Resources/app/administration/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal/index.ts
larskemper: just a simple validation check before sending a request every time

PR #120, opened 2026-01-26
refactor: logging service

larskemper:
refactor of `src/Migration/Logging/LoggingService.php`

#### auto save

- added max buffer size and auto flush/save when max size has been reached
- also saves when service get deconstructed

#### simplification

- removed context passing and use default shopware context
- rename: `addLogEntry()` & `saveLogging()` => `log()` & `flush()`
- auto save => you don't have to save all the time
- chain `flush()` => $this->loggingService->log(...)->flush() instead of separat calls


src/Migration/Logging/LoggingService.php
```diff
@@ -27,33 +36,52 @@ public function __construct(
...
-            $this->logging = [];
+            $this->reset();
         }
     }
 
-    public function addLogEntry(MigrationLogEntry $logEntry): void
+    public function log(MigrationLogEntry $logEntry): self
     {
-        $this->logging[] = [
+        $key = $this->generateKey($logEntry);
```
larskemper: Not sure if we really profit from the unique key (avoiding duplicate logs). What do you think?

jozsefdamokos: I dont see the need 🤷

src/Migration/Media/Processor/HttpDownloadServiceBase.php
```diff
@@ -78,14 +78,13 @@ public function process(MigrationContextInterface $migrationContext, Context $co
...
-            $this->loggingService->addLogEntry(
+            $this->loggingService->log(
                 MigrationLogBuilder::fromMigrationContext($migrationContext)
                     ->withExceptionMessage($exception->getMessage())
                     ->withExceptionTrace($exception->getTrace())
                     ->withEntityName(MediaDefinition::ENTITY_NAME)
                     ->build(ExceptionRunLog::class)
-            );
-            $this->loggingService->saveLogging($context);
+            )->flush();
```
jozsefdamokos: I see some places where `save` was called before you add the `flush` and some places you dont. What is the reason behind it? From what i understand the flush would not be needed at all, it will be called internally.

jozsefdamokos (approved):

src/Migration/Media/Processor/HttpDownloadServiceBase.php
larskemper: i first thought that we might not want to fully really on the auto flush, because e.g. "in cause of an exception can we make sure that the deconstruct/reset will work?", but thats also true for the manual saving.

So I removed all these calls 👍

src/Profile/Shopware/Gateway/Api/Reader/TableCountReader.php
vintagesucks: Is `Context $context` still needed here?

src/Migration/Logging/LoggingService.php
```diff
@@ -12,13 +12,21 @@
...
 use Shopware\Core\Framework\Log\Package;
 use SwagMigrationAssistant\Migration\Logging\Log\Builder\MigrationLogEntry;
+use Symfony\Contracts\Service\ResetInterface;
 
 #[Package('fundamentals@after-sales')]
-class LoggingService implements LoggingServiceInterface
+class LoggingService implements LoggingServiceInterface, ResetInterface
 {
-    protected array $logging = [];
+    final public const BUFFER_SIZE = 50;
```
vintagesucks, line 20: should we make this configurable?

vintagesucks (approved):
nice one 👍

src/Migration/Logging/LoggingService.php
larskemper: we could, but as basically everything else is also not configurable currently, i would maybe think about adding this in a bigger initiative, so also for other configs in one ticket.

larskemper: created a ticket for this: https://github.com/shopware/shopware/issues/14598. We can discuss this and decide then

PR #121, opened 2026-01-27
fix: missing migration state checks

larskemper:
some validation checks where missing e.g. user could remove disabled state of "continue" migration button (error-resolution) and proceed without providing fixes.


vintagesucks (approved):

tests/unit/Migration/Run/MigrationStepTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        $step->assertOneOf(...$abortableSteps);
+    }
+
+    /**
+     * @return iterable<string, array{MigrationStep, bool}>
+     */
+    public static function provideAbortableSteps(): iterable
+    {
+        yield 'FETCHING is abortable' => [MigrationStep::FETCHING, true];
```
jozsefdamokos: We should add keys I think

jozsefdamokos (approved):

PR #126, opened 2026-01-29
test: migration ats

larskemper:
resolves https://github.com/shopware/shopware/issues/12706

adds migration ats & visual tests + nightly run


.github/workflows/acceptance.yaml
```diff
@@ -56,7 +60,7 @@ jobs:
         run: |
           composer run build:js:admin
           cd custom/plugins/${{ github.event.repository.name }}/tests
-          mysql -uroot < _fixtures/database/sw55.sql
```
jozsefdamokos, line 59: This is now a much smaller database from what i can tell. Does it still contain all the same types of data and errors that the previous database had? The test was useful before to identify bugs (for example if the number of migrated entities have changed because of a change we made). Now that is much simpler Im not sure it stays just as useful. What do you think?

jozsefdamokos (dismissed):

.github/workflows/acceptance.yaml
larskemper: yep good one, will look at that

MalteJanz: Yes I would also suggest we stick to the old DB for now, especially to test our feature the same way that it was tested on `trunk`.

But in case you also want to test your 5.7 DB: Maybe add a second job (e.g. make this CI job a matrix where the SW5 DB changes)

tests/acceptance/.gitignore
```diff
@@ -1,2 +1,7 @@
 playwright-report/
 test-results/
+
+snapshots/**
+!snapshots/*/
+!snapshots/*/linux/
+!snapshots/*/linux/**
```
MalteJanz, line 8: Why are these snapshot files ignored here? The whole purpose of them is that we commit them and that they are only intentionally changed (e.g. if the log file changes intentionally, instead of accidentally by a change).

MalteJanz (changes requested):

tests/acceptance/.gitignore
larskemper: they are not ignored, therefore the `!`. When generating locally for mac and win system it will create different folders (darwin, win). So that nobody accidentally commit these we ignore every snapshot folder except these ones with linux folders and its childs

.github/workflows/acceptance.yaml
larskemper: reverted and use the old one again

tests/acceptance/.gitignore
MalteJanz: Ah makes sense sorry. A code comment about that would be helpful though (nitpick)

.github/workflows/acceptance.yaml
```diff
@@ -56,7 +60,7 @@ jobs:
...
+          mysql -uroot < _fixtures/database/shopware55corrupted.sql
```
MalteJanz, line 63: Shouldn't this be `shopware55.sql`? Or ask differently what is `shopware55corrupted.sql`?

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
```diff
@@ -0,0 +1,3796 @@
...
+ID: [uuid]
+Level: error
+Code: SWAG_MIGRATION_WRITE_EXCEPTION
+Profile name: shopware55
+Gateway name: local
+Created at: [timestamp]
+Entity: order
+Exception message: There are 1 error(s) while writing data.
+
+1. [/1/orderCustomer/email] This value should not be blank.
```
MalteJanz, line 2363: Unrelated to this PR, but we should look into why this wasn't catched by the validation 👀

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
```diff
@@ -0,0 +1,3796 @@
...
+        "interval": 0.01,
+        "roundForNet": true
+    },
+    "currencyFactor": 1,
+    "salesChannelId": "[uuid]",
+    "billingAddressId": "[uuid]",
+    "primaryOrderDeliveryId": "[uuid]",
+    "primaryOrderTransactionId": "[uuid]"
+}
+Exception trace (JSON):
```
MalteJanz, line 3540: Again unrelated, but can we maybe limit the trace output to the first 10 entries in the log file. This list is huge otherwise 😅

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            const finalBuffer = Buffer.concat(buffers);
+            let logString = finalBuffer.toString();
+
+            // remove media warning logs
+            logString = logString.replaceAll(
+                /----- Log Entry #[0-9]+ -----\nID: [0-9a-f]{32}\nLevel: warning\nCode: SWAG_MIGRATION_MEDIA_FILE_MISSING(.|\n)+?\n\n/g,
+                '',
+            );
```
MalteJanz: Not sure if this is still strictly needed. In the past we needed it because the ordering of these logs was not deterministic (because the download ran in parallel). But since then we switched to serial processing, so maybe the order is deterministic now and we can leave this in the compared / tested log file.

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+            await expect(page.locator('.swag-migration-error-resolution-step__card-table-count-icon')).toHaveCount(logCount);
+
+            await page.locator('.swag-migration-error-resolution-step__header-buttons-continue').click();
+            await waitForLoaders(page);
+
+            await page.locator('.swag-migration-error-resolution-step__continue-modal-confirm').click();
+            await waitForLoaders(page);
+        });
+
+        await test.step('Finish migration', async () => {
```
MalteJanz, line 186: Would be great if you could add this step as well again, to check that the entity counts are exactly the same and didn't change in our feature branch compared to trunk
https://github.com/shopware/SwagMigrationAssistant/blob/aa9189f7e90a13ef64094399b6ec1ef0ee8642f1/tests/acceptance/tests/MigrationByUiFlow.spec.ts#L114-L134

And also during future changes we do

.github/workflows/acceptance.yaml
larskemper: Just a db with invalid data, so we can also test the error resolution deterministically in the ats

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
larskemper: right, currently the missing field check is not enabled for sub child entities, cause of the id problem. But I guess we could easily enable it for simple scalar field (no fk field, ...)

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+
+            await page.getByTitle('Data selection').click();
+            await waitForLoaders(page);
+
+            await page.locator('.swag-migration-data-selector input[type="checkbox"]').first().check();
+            await waitForLoaders(page);
+
+            const checkboxes = page.locator('.swag-migration-data-selector input[type="checkbox"]');
+
+            await expect(checkboxes).toHaveCount(await checkboxes.count());
```
vintagesucks: Am I missing something or will this always pass? What does this test? 😅

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
larskemper: https://github.com/shopware/shopware/issues/14860

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
MalteJanz, line 1: Should this file be even bigger now? (I can't review it on GitHub anymore and would need to checkout the branch locally 🙈 )

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+            await EntityCounter.checkEntityCount('swag_migration_logging', 698);
+
+            await EntityCounter.checkEntityCount('product', 427);
+            await EntityCounter.checkEntityCount('product_review', 2);
+            await EntityCounter.checkEntityCount('category', 63);
+            await EntityCounter.checkEntityCount('property_group', 14);
+            await EntityCounter.checkEntityCount('property_group_option', 93);
+            await EntityCounter.checkEntityCount('product_manufacturer', 14);
+
+            await EntityCounter.checkEntityCount('order', 1);
```
MalteJanz, line 229: Is it intentional that there is one less order migrated than in `trunk`?

https://github.com/shopware/SwagMigrationAssistant/blob/1704a1b42bb035fd25bafbbd5a727c60868118c5/tests/acceptance/tests/MigrationByUiFlow.spec.ts#L114

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+            await EntityCounter.checkEntityCount('customer', 3);
+
+            await EntityCounter.checkEntityCount('cms_page', 10);
+            await EntityCounter.checkEntityCount('media', 595);
+            await EntityCounter.checkEntityCount('media_folder', 24);
+            await EntityCounter.checkEntityCount('document', 4);
```
MalteJanz, line 235: see above:
https://github.com/shopware/SwagMigrationAssistant/blob/1704a1b42bb035fd25bafbbd5a727c60868118c5/tests/acceptance/tests/MigrationByUiFlow.spec.ts#L120

MalteJanz (commented):
The changes look mostly good to me now, nice work 👍 .

But maybe trigger the acceptance pipeline a few times, just to make sure it's not flaky, before merging it.

tests/acceptance/tests/MigrationTest.spec.ts
larskemper: it's cause of this one i guess https://github.com/shopware/SwagMigrationAssistant/pull/126#discussion_r2774524580. The order is not written cause of the write exception. Most likely also something wrong in the Converter... (should soft escape the ordercustomer when invalid, so the order can be written anyway)

tests/acceptance/tests/MigrationTest.spec.ts
larskemper: same here, the documents should belong to the order that is not written...

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
larskemper: compared the exceptions, most of them are exactly the same as before. The size grew, cause we write a lot more data to it with the new logging information we have

MalteJanz (approved):

vintagesucks (approved):

PR #133, opened 2026-02-13
fix: restore converter checks

larskemper:
resolves https://github.com/shopware/shopware/issues/14778


src/Profile/Shopware/Converter/CrossSellingConverter.php
```diff
@@ -80,17 +86,15 @@ public function convert(array $data, Context $context, MigrationContextInterface
...
         $converted['type'] = 'productList';
         $converted['active'] = true;
+        $converted['productId'] = $sourceProductId;
         $converted['assignedProducts'] = [
             [
-                'id' => $relationMapping['entityId'] ?? null,
-                'position' => $data['position'] ?? null,
-                'productId' => $relatedProductMapping['entityId'] ?? null,
+                'id' => $relationMapping['entityId'],
+                'position' => $data['position'],
```
MalteJanz: Why no `null` fallback here anymore?

src/Profile/Shopware/Converter/NewsletterRecipientConverter.php
```diff
@@ -161,21 +168,44 @@ protected function getSalutation(string $salutation, MigrationContextInterface $
...
-                $this->connectionId,
-                DefaultEntities::SALES_CHANNEL,
-                $data['shopId'],
-                $this->context
-            );
-        }
+        $salesChannelMapping = $this->mappingService->getMapping(
+            $this->connectionId,
+            DefaultEntities::SALES_CHANNEL,
+            $data['shopId'],
```
MalteJanz, line 177: array key `$data['shopId']` might not exists. Was checked before with `isset`. You could do an fallback:

```suggestion
            $data['shopId'] ?? '',
```

because in practice that shouldn't find a mapping

src/Profile/Shopware/Converter/TranslationConverter.php
```diff
@@ -133,10 +147,11 @@ protected function createProductTranslation(array &$data): ConvertStruct
                     ->build(ConvertAssociationMissingLog::class)
             );
 
-            return new ConvertStruct(null, $sourceData);
+            $product['id'] = null;
```
MalteJanz: Does it make sense to migrate a product translation without associated product (e.g. is this fixable with error resolution)?

My POV: leave the early return in here and just skip migrating that translation if we can't find the associated product

src/Profile/Shopware/Gateway/Local/Reader/SeoUrlReader.php
```diff
@@ -121,7 +121,7 @@ private function isRouterToLower(MigrationContextInterface $migrationContext)
         }
 
         /** @phpstan-ignore shopware.unserializeUsage */
-        return (bool) \unserialize($useUrlToLower, ['allowed_classes' => false]);
+        return (bool) \unserialize($useUrlToLower);
```
MalteJanz: I added that in my last PR and we should keep it 😉

MalteJanz (changes requested):

ennasus4sun (approved):

src/Profile/Shopware/Gateway/Local/Reader/SeoUrlReader.php
larskemper: ups 🤣

src/Profile/Shopware/Converter/CrossSellingConverter.php
larskemper: change default handling of `productId`, missed the others 👍

MalteJanz (approved):

PR #135, opened 2026-02-16
chore: align cs fixer rules

larskemper:
aligned cs fixer rules with platform: https://github.com/shopware/shopware/blob/trunk/.php-cs-fixer.dist.php


larskemper:
what would you think about a rule like this ?: 

```php
'blank_line_before_statement' => [
    'statements' => [
        'return',
        'throw',
        'continue',
        'break',
        'if', // optional
        'foreach', // optional
        'for', // optional
        'while', // optional
        'switch', // optional
        'try', // optional
    ],
],
```

so adding a empty line before important keywords, makes the code much easier to read in my opinion.

src/Migration/Run/MigrationStep.php
```diff
@@ -44,6 +34,15 @@ enum MigrationStep: string
     case FINISHED = 'finished';
 
     case ABORTED = 'aborted';
+    final public const MANUAL_STEPS = [
```
vintagesucks, line 38:
```suggestion

    final public const MANUAL_STEPS = [
```

Not in favor of the `blank_line_before_statement` rule from your comment question, but a blank line would not hurt here 😅

vintagesucks (approved):

src/Migration/Run/MigrationStep.php
larskemper: ok 🤣. weird i did just run the cs

MalteJanz (approved):
LGTM 👍 . Regarding your comment: I would prefer staying consistent with whatever our shopware core is doing.

Also in my POV empty lines mostly depend on context and are rather subjective. E.g. if I have a `throw` as the only statement inside an `if` I most likely don't want an empty line before that, but if the `if` contains other statements before, it makes it easier to read to visually separate certain groups of things.

PR #137, opened 2026-02-17
fix: validate new connection before storing

ennasus4sun:

Currently, when you **create a new connection**, it is already stored in the database before validation.
wizard:
profile informations (already stored into DB) > credentials (validates that connection to source system is possible) > success || error

- new flag `isNewConnection`

**profile "page"**
- creates now a new Connection-Entity and store it in memory

**credentials "page"**
- `checks for isNewConnection === true` (if not, the old implementation is used (`doConnectionCheck` > `updateConnectionCredentials`); another possibility would be "edit connection" )
- calls new backend endpoint `createNewConnection`
- uses a transaction to create a record in the database if all validations (check profilename, is it possible to connect to source system and fingerprint) are valid
<br /><br /><br />
closes: https://github.com/shopware/shopware/issues/14001


src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard-page-credentials-error/swag-migration-wizard-page-credentials-error.html.twig
```diff
@@ -29,7 +29,7 @@
                         <mt-banner
                             v-if="errorMessageHintSnippet"
                             class="swag-migration-wizard__description-hint"
-                            :variant="errorMessageHintSnippetType"
+                            :variant="errorMessageHintType"
```
ennasus4sun, line 32: just FYI: property has a different name: 
https://github.com/shopware/SwagMigrationAssistant/blob/feature/migration-logging-refactor/src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard-page-credentials-error/index.ts#L24

tests/Jest/src/module/swag-migration/page/wizard/swag-migration-wizard.spec.js
```diff
@@ -0,0 +1,513 @@
...
+            await wrapper.find('.swag-migration-wizard__footer button.mt-button--primary').trigger('click');
+
+            await flushPromises();
+
+            expect(router.currentRoute.value.name).toBe('swag.migration.wizard.credentials');
+
+            // if router route name does not match with this->currentRoute
+            // trigger matchCurrentRoute() execution to update this.currentRoute
+            if (wrapper.vm.currentRoute.name !== router.currentRoute.value.name) {
+                wrapper.vm.matchCurrentRoute(false);
```
ennasus4sun: needs to be investigated how this can be avoided (internal slack: https://shopware-ag.slack.com/archives/C0937LSK284/p1771408033169819?thread_ts=1771338466.186999&cid=C0937LSK284)

ennasus4sun: moved code into helper method (outdated = not visible any more), because of that paste it here:
```
async function syncCurrentRoute(wrapper, routerCurrentRouteName) {
    if (wrapper.vm.currentRoute.name !== routerCurrentRouteName) {
        wrapper.vm.matchCurrentRoute(false);
        await flushPromises();
    }
}
```

src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard/swag-migration-wizard.html.twig
```diff
@@ -22,7 +22,6 @@
                         :is="Component"
                         v-show="!isLoading && !childIsLoading && !storesInitializing"
                         :current-connection-id="connection?.id"
-                        :profile-information-component="profileInformationComponent"
```
ennasus4sun, line 25: I assume this was just forgotten to be removed 🤷🏻 
for me it looks similar to the credentialsComponent logic (https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard/index.js#L172-L178 > loads different forms for different combinations of sw5, sw6, api and local ) which would not make much sense for the profile page.

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
         return new JsonResponse(\array_values($this->dataSelectionRegistry->getDataSelections($migrationContext, $environmentInformation)->getElements()));
     }
 
+    #[Route(
+        path: '/api/_action/migration/create-new-connection',
+        name: 'api.admin.migration.create-new-connection',
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']],
```
vintagesucks:
```suggestion
        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.editor']],
```

DennisGarding (approved):

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
...
+    public function createNewConnection(Request $request, Context $context): JsonResponse
+    {
+        $id = $request->request->getAlnum('connectionId');
+        $connectionName = (string) $request->request->get('connectionName');
+        $profileName = (string) $request->request->get('profileName');
+        $gatewayName = (string) $request->request->get('gatewayName');
+        $credentialFields = $request->request->all('credentialFields');
+
+        if ($id === '') {
+            throw RoutingException::missingRequestParameter('id');
```
vintagesucks:
```suggestion
            throw RoutingException::missingRequestParameter('connectionId');
```

vintagesucks (approved):

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
...
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('name', $connectionName));
+
+        $existingConnection = $this->migrationConnectionRepo->search($criteria, $context)->getEntities()->first();
+
+        if ($existingConnection !== null) {
+            throw MigrationException::connectionNameNotUnique();
+        }
+
+        $jsonResponse = $this->connection->transactional(
```
jozsefdamokos: This controller method is quite large. How about we extract the transaction part into a service`Migration\Connection\Service\ConnectionService::createNewConnection`? The integration tests for the controller can stay so not much extra work

jozsefdamokos (approved):

src/Controller/StatusController.php
ennasus4sun: thank you @vintagesucks.
I did not had a look regarding the permissions.
When reading the [permissions definitions](https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Resources/app/administration/src/module/swag-migration/index.js)  it feels for me like it should have `creator` permissions...
any other thoughts?

vintagesucks: I think you are correct 👍

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
...
+                    }
+                );
+
+                $connection = $this->migrationConnectionRepo->search(new Criteria([$id]), $context)->getEntities()->first();
+
+                if ($connection === null) {
+                    throw MigrationException::noConnectionFound();
+                }
+
+                $migrationContext = $this->migrationContextFactory->createByConnection($connection);
```
MalteJanz: Why is it needed to first write it to DB and immediately fetch the same data again?

I would suggest something like this:

```suggestion
                $connectionData = [
                    'id' => $id,
                    'name' => $connectionName,
                    'profileName' => $profileName,
                    'gatewayName' => $gatewayName,
                    'credentialFields' => $credentialFields,
                ];

                $context->scope(
                    MigrationContext::SOURCE_CONTEXT,
                    function (Context $scopedContext) use ($connectionData): void {
                        $this->migrationConnectionRepo->create([$connectionData], $scopedContext);
                    }
                );

                $connectionEntity = new SwagMigrationConnectionEntity();
                $connectionEntity->assign($connectionData);
                $migrationContext = $this->migrationContextFactory->createByConnection($connectionEntity);
```

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
...
+                    $id,
+                );
+
+                if ($hasDuplicate) {
+                    throw MigrationException::duplicateSourceConnection();
+                }
+
+                $updateData = [
+                    'id' => $id,
+                    'credentialFields' => $credentialFields,
```
MalteJanz: why write the credentials again here? There were written a second ago and as far as I can see didn't change? 🤔

src/Controller/StatusController.php
MalteJanz: Why is the transaction even needed in the first place? Couldn't you just create a connection entity in memory, do the connection check (environment call) + fingerprint, and only then write to DB before returning the environment information?

src/Controller/StatusController.php
```diff
@@ -233,6 +236,121 @@ public function getDataSelection(Request $request, Context $context): JsonRespon
...
+                }
+
+                $context->scope(
+                    MigrationContext::SOURCE_CONTEXT,
+                    function (Context $scopedContext) use ($updateData): void {
+                        $this->migrationConnectionRepo->update([
+                            $updateData,
+                        ], $scopedContext);
+                    }
+                );
```
MalteJanz: If you don't update the credentials, you don't need the extra context scope here.

Also you essentially never need to update (write to DB twice) if you refactor all of this without a DB transaction and could only write everything once at the end.

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -73,6 +73,39 @@ export default class MigrationApiService extends ApiService {
         };
     }
 
+    async createNewConnection(
+        connectionId: string,
+        connectionName: string,
+        profileName: string,
+        gatewayName: string,
+        credentialFields: Record<string, MigrationCredentials>,
```
MalteJanz, line 81: This record type doesn't make sense. I would suggest
```suggestion
        credentialFields: Record<string, string|number|boolean>,
```

Nitpick because it wasn't part of your PR: the type `MigrationCredentials` is too strict, because it only considers the schema of the SW5 API connection, but not other possible profile + gateway combinations. We should change it to allow for arbitrary credentials

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -73,6 +73,39 @@ export default class MigrationApiService extends ApiService {
...
+        additionalHeaders: AdditionalHeaders = {},
+    ): Promise<ApiResponse<unknown>> {
```
MalteJanz, line 83: Doesn't the API (success) response always follows the same schema so we can do something similar to the `checkConnection` method?

```suggestion
    ): Promise<ApiResponse<MigrationEnvironmentInformation>> {
```

Nitpick: again unrelated to this PR but I'm not sure if the TypeScript typechecker is running correctly on these files, because some of the return types look wrong to me (maybe it's also because all of these `@ts-ignore` comments, which we should try to avoid, other api services in the core also work fine without them...) 🤔

src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard/index.ts
```diff
@@ -293,13 +295,28 @@ export default Shopware.Component.wrapComponentConfig({
...
-                    await this.migrationApiService.updateConnectionCredentials(
+                if (this.isNewConnection) {
+                    const environmentInformation = await this.migrationApiService.createNewConnection(
                         this.connection.id,
+                        this.connectionName,
+                        this.selectedProfile.profile,
+                        this.selectedProfile.gateway,
                         this.connection.credentialFields,
                     );
+
```
MalteJanz, line 307: After the new connection got successfully created (in the method above), we should also set the flag `isNewConnection` to `false` again.

Otherwise this edge case isn't covered: one might go back and change the connection again, and instead of updating it it tries to create another one with the same ID, which will always fail 🤔

src/Resources/app/administration/src/module/swag-migration/page/wizard/swag-migration-wizard/swag-migration-wizard.html.twig
MalteJanz: Yes that page was used to show different information based on profile + gateway combination. But I guess Lars removed that and moved the information into the credentials pages themselves.

So yes this is dead code and can be removed, but looks like Lars forgot to remove some of these information pages too. Would be nice if you can remove them as well (look through the profile folder), like this one:
https://github.com/shopware/SwagMigrationAssistant/blob/feature/migration-logging-refactor/src/Resources/app/administration/src/module/swag-migration/profile/shopware54/api/swag-migration-profile-shopware54-api-page-information/index.ts

tests/Jest/src/module/swag-migration/page/wizard/swag-migration-wizard.spec.js
```diff
@@ -0,0 +1,507 @@
...
+            component: () => Shopware.Component.build('swag-migration-wizard-page-credentials-success'),
+        },
+        {
+            path: '/credentials/error',
+            name: 'swag.migration.wizard.credentialsError',
+            component: () => Shopware.Component.build('swag-migration-wizard-page-credentials-error'),
+        },
+    ],
+    history: createWebHashHistory(),
+});
```
MalteJanz: According to the vue test utils docs each test case needs a new instance of a router:
https://test-utils.vuejs.org/guide/advanced/vue-router.html#Using-a-Real-Router

> in case you prefer to keep using a real router, keep in mind that each test should use it's own instance of the router like so:
```ts
let router
beforeEach(async () => {
  router = createRouter({
    history: createWebHistory(),
    routes: routes
  })
})
```

I guess because the real router contains internal state (like the history) and to prevent that leaking into different test cases and causing strange behavior (e.g. when navigating back). So it would be good practice to follow this even though the current implementation might work fine for the current test cases.

MalteJanz (changes requested):

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: I will leave this like it is for now and investigate in a different PR after this is merged.

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: I will leave this like it is for now and investigate in a different PR after this is merged.

tests/Jest/src/module/swag-migration/index.spec.js
MalteJanz, line 1: Fyi: I removed this test because I don't see any value in testing our module declaration (which boils down testing a plain old data object / configuration without any logic). In practice it will fail any time a component is added / removed, which I find is only annoying and slowing us down.

Let me know if you think differently and I should add this back

MalteJanz (approved):

src/Controller/StatusController.php
MalteJanz: I refactored the logic into a `MigrationConnectionFactory` and also updated the other method in the StatusController to reuse the same logic. Please take another look 🙂

jozsefdamokos (approved):

src/Migration/Connection/MigrationConnectionFactory.php
```diff
@@ -0,0 +1,130 @@
...
+ */
+#[Package('fundamentals@after-sales')]
+readonly class MigrationConnectionFactory
+{
+    /**
+     * @param EntityRepository<SwagMigrationConnectionCollection> $migrationConnectionRepo
+     */
+    public function __construct(
+        private EntityRepository $migrationConnectionRepo,
+        private MigrationContextFactory $migrationContextFactory,
```
vintagesucks:
```suggestion
        private MigrationContextFactoryInterface $migrationContextFactory,
```

src/Migration/Connection/MigrationConnectionFactory.php
```diff
@@ -0,0 +1,130 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\NotFilter;
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Exception\MigrationException;
+use SwagMigrationAssistant\Migration\Connection\Fingerprint\MigrationFingerprintServiceInterface;
+use SwagMigrationAssistant\Migration\EnvironmentInformation;
+use SwagMigrationAssistant\Migration\MigrationContext;
+use SwagMigrationAssistant\Migration\MigrationContextFactory;
```
vintagesucks:
```suggestion
use SwagMigrationAssistant\Migration\MigrationContextFactoryInterface;
```

see below

src/Migration/Connection/MigrationConnectionFactory.php
vintagesucks:
```suggestion
        private MigrationContextFactoryInterface $migrationContextFactory,
```

vintagesucks (approved):

PR #141, opened 2026-02-24
fix: skip app-owned custom field sets

vintagesucks:
resolves shopware/shopware#14656

Skips custom field sets with an `app_id` during migration to avoid FK constraint violations. Expands `exception_message` column to `LONGTEXT` so long DB error messages are no longer silently dropped from the migration log.


src/Core/Migration/Migration1754897550AddFieldsToMigrationLogs.php
```diff
@@ -39,7 +39,7 @@ class Migration1754897550AddFieldsToMigrationLogs extends MigrationStep
         'field_source_path' => 'VARCHAR(255) NULL',
         'source_data' => 'JSON NULL',
         'converted_data' => 'JSON NULL',
-        'exception_message' => 'VARCHAR(255) NULL',
+        'exception_message' => 'LONGTEXT NULL',
```
vintagesucks, line 42: This [uncovered](https://github.com/shopware/SwagMigrationAssistant/actions/runs/22347487000/job/64665754598?pr=141) another previously-swallowed exception:

```
----- Log Entry #8 -----
      ID: [uuid]
      Level: error
      Code: SWAG_MIGRATION_RUN_EXCEPTION
      Profile name: shopware55
      Gateway name: local
      Created at: [timestamp]
      Entity: product
      Exception message: SwagMigrationAssistant\Profile\Shopware\Converter\ProductConverter::getPrice(): Argument #1 ($priceData) must be of type array, null given, called in /home/runner/work/SwagMigrationAssistant/SwagMigrationAssistant/custom/plugins/SwagMigrationAssistant/src/Profile/Shopware/Converter/ProductConverter.php on line 383
```

I added this log entry to our `migration-log-sw5.txt` snapshot for now.

jozsefdamokos: Should we check for it in the converter and create an `MigrationValidationRequiredFieldMissingLog` instead of the exception?

jozsefdamokos (approved):

MalteJanz (approved):

src/Core/Migration/Migration1754897550AddFieldsToMigrationLogs.php
vintagesucks: I tried to do this, but it leads to ATS failures. I added `editor` field handling, but ATS fails due to a validation error preventing error resolution from completing, any idea how to solve this?

image

MalteJanz: Looks like it uncovered another shortcoming of our current approach: How would the user specify the correct UUID there (the one of the default currency of the shop)? We might need to implement a special Vue component for these price fields, because manually editing them in JSON is not very user friendly, especially if you need to figure out UUIDs yourself 🙈 .

Nevertheless in this case I would suggest ignoring the error in this PR and creating a follow up issue to deal with it

vintagesucks: https://github.com/shopware/shopware/issues/15204

PR #142, opened 2026-02-26
fix: strip exception traces from snapshot test

vintagesucks:
see https://github.com/shopware/SwagMigrationAssistant/pull/41#issuecomment-3966990828


jozsefdamokos (approved):

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -273,6 +273,12 @@ test.describe('Migration Tests @migration @visual', () => {
             // replace ids
             logString = logString.replaceAll(/[0-9a-f]{32}/g, '[uuid]');
 
+            // remove exception traces
+            logString = logString.replaceAll(
+                /Exception trace \(JSON\):\n\[[\s\S]*?\]\n/g,
```
MalteJanz, line 278: This regex seems to work fine for our case, I was just thinking if it is possible that further JSON arrays appear inside the trace, because that would then only match to the first `]` right? 🤔

MalteJanz (approved):

MalteJanz:
@vintagesucks I just went ahead and merged it, I hope this was fine ✌️

PR #132, opened 2026-02-10
chore: benchmark and performance improvements

MalteJanz:
### Environment

- MacBook M4 Pro
- local development setup (Apache + PHP8.3.30 + MySQL8.4.8)
- SW5 with demo data and Connector plugin
- SW6 using Migration-Assistant and SW5.7 profile + API gateway
- admin worker disabled, running one local CLI message queue worker

### Baseline

Profiled the fetch + convert (+ validation) of a single `MigrationProcessMessage` that contains:
- entity: `product`
- offset: `0`
- limit: `100`

#### `trunk`

https://blackfire.io/profiles/44219502-e201-4da3-a919-cc18da079032/graph

image

#### `feature/migration-logging-refactor`

https://blackfire.io/profiles/7f7465ab-62c5-423f-9e23-fda9b286be0d/graph

image

### Results

`chore/benchmark-and-performance-improvements`:

https://blackfire.io/profiles/c5bf1276-c579-4326-925d-1fff7d2ea18c/graph

image

| Branch | Wall time | Improvement(time) | Memory | Improvement(Mem) |
|--------|--------|--------|--------|--------|
| `trunk` | 1.541s | - | 31.02MB | - |
| `feature/migration-logging-refactor` |2.577s | baseline | 64.51MB | baseline |
|`chore/benchmark-and-performance-improvements` |1.907s | 26% |37.07MB | 43% |

### Disclaimer

during my benchmarks the IO time fluctuated quite a bit (e.g. sometimes the DB took longer or was faster), so don't fully trust the wall time improvements in absolute or relative numbers. Better would be just inspecting the validate function in Blackfire and just comparing that.

### How to test / reproduce

Install the blackfire php SDK in shopware platform folder with:

```bash
composer require blackfire/php-sdk
```

And have the blackfire agent running + correctly configured.

Add the following below this line:
https://github.com/shopware/SwagMigrationAssistant/blob/45d1f464390375942eea145ca366d380b5e262eb/src/Migration/MessageQueue/Handler/Processor/FetchingProcessor.php#L63

```php
        if ($progress->getCurrentEntity() === 'product' && $progress->getCurrentEntityProgress() === 0) {
            $blackfire = new \Blackfire\Client();
            $config = new \Blackfire\Profile\Configuration();
            $config->setTitle('Fetching ' . $progress->getCurrentEntity() . ' ' . $progress->getCurrentEntityProgress() . ' / ' . $totalCountOfCurrentEntity . ' limit ' . $migrationContext->getLimit());
[...]

PR #144, opened 2026-02-27
perf: improve mapping lookup performance

MalteJanz:
## Disclaimer

**This is just an experiment / PoC for now (created on a circle day). It's not yet decided if we move forward with this refactor.**

## Context

- In our Migration-Assistant we have `Converter` classes which map data from the source system to the SW6 schema
- This process runs in batches of 100 entities, e.g. 100 products and for each one the corresponding `Converter` `convert` method is called
- During the conversion, the `Converter` has to map foreign keys in lots of places (e.g. `manufacturerId` of `42` in SW5 to the correct SW6 UUID). This is where the `MappingService` comes in
- Right now it's doing one DB query for every mapping lookup, but caches the results in memory. The chance of a cache miss is assumed to be rather high
- Many small DB queries are slow and it's often called the [N+1 Query problem](https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it)
- It would be nice to fetch all necessary Mappings in a single DB query to improve the performance significantly
- And doing so without making the converter logic too complex
  - e.g. we also had the optimization to "remember" all used mappings for an entity and storing that, so for future migrations these are preloaded. This was implemented in a manual way and is quite easy to miss and not benefit from. I would like to get rid of this extra complexity that only brings value for the second migration onward of the same entity and instead provide a solution that works all the time and is rather simple.

## Baseline benchmark

I explained the environment of these benchmarks [in a previous PR](https://github.com/shopware/SwagMigrationAssistant/pull/132#issue-3920560389) so I only focus on the results here:

### Perfect environment
Let's start with looking at a usual `product` convert batch of 100 entities:
https://blackfire.io/profiles/37d53432-4c24-4f51-b79e-507449c25ee6/graph

The `SwagMigrationAssistant\Migration\Mapping\MappingService::getMapping` method:
- has a total wall time of `477ms` and is responsible for `23,67%` of the total wall time of the process
- is called `3047` times
- reached out to the DB `1263` times

But you have to keep in mind that in my dev environment, the DB is running on the same machine than the message worker PHP process. Means my network latency is near zero.

Let's see how it looks if we add a small delay of `1ms` every time we reach the DB in the mapping method (by adding a `usleep(1000)` call there for the non cached case.

### Simulating production

1ms is just an assumption, ChatGPT told me this:
- Same host RTT: ~0.01–0.1 ms
- Same availibity zone in AWS: ~0.2–0.8 ms (already ~5–20x slower than local loopback)
- Same region but different zone in AWS: ~1–3 ms (~10–100x higher than local)

With that 1ms I get this result:
https://blackfire.io/profiles/57a1ea2b-0332-4584-a320-47d89942d400/graph

Notice how much different the situation looks now for the `getMapping` method:
- has a total wall time of `2,28s` and is responsible for `50,36%` of the total wall time of the process
- is still called the same `3047` times
- reached out to the DB the same `1263` times but also called `usleep(1000)` the same number of times for simulating the network latency.

This makes it more obvious why this N+1 Query problem is such a big deal for all applications that talk to the DB in production.

## Proposed solution

To keep the converter logic mostly untouched and simple to read, I would propose a concept similar to asynchronous computing (like Promises from the JS world).

The idea is simple:
- Each `getMapping` call results in a placeholder / promise at first, but registers a task to lookup a certain mapping from the DB
- At the end we can fetch all mappings from the DB in a single query
- And replace the placeholders / fullfil the promises with the real data

One way of implementing this would be to store absolute paths inside the nested `$converted` array to remember which string needs to be updated, but that would be really cumbersome and error prone to use.

Fortunately PHP is quite powerful and supports [variable references / aliases](https://www.php.net/manual/en/language.references.php). I tried to experiment using them a little like pointers from lower level languages and to some degree that seems to work (see [mapping-experiment.php](https://github.com/shopware/SwagMigrationAssistant/blob/perf/improve-mapping-lookup-performance/mapping-experiment.php))

## Result benchmark

https://blackfire.io/profiles/ff569f15-3518-47e3-b7e4-13fb5806f736/graph

The `SwagMigrationAssistant\Migration\Mapping\MappingServiceV2::resolvePromises`method:
- has a total wall time of `257ms` and is responsible for `16,39%` of the total wall time of the process
- is called `1` time
- reached out to the DB `2` times (one time fetch existing + one time write new mappings)

Below a summary of the total wall time and memory of the entire process (instead of just looking at mapping service calls):

| Scenario                                      | Time (s) | Memory (MB) | Time Δ vs Baseline | Memory Δ vs Baseline | Time Δ vs Latency Baseline |
|-----------------------------------------------|----------|-------------|---------------------|----------------------|----------------------------|
| Baseline                                     | 2.02     | 37.6        | —                   | —                    | -55.4%                     |
| Baseline (+1ms simulated DB network latency) | 4.53     | 39.0        | +124.3%             | +3.7%                | —                          |
| This change                                      | 1.57     | 25.0        | -22.3%              | -33.5%               | -65.3%                     |

So the expected wall time improvement is likely anywhere between 22% and 65% (or even more when having worse DB network latency). When we apply this refactor to all converters and you consider that real migrations often run hours, the time savings are quite significant.

Downsides:
- quite some work to refactor all converters, but at the same time the mappings calls themselves are simpler (fewer arguments, just return the mapping UUID instead of an array)
- it's easier to introduce errors (e.g. forgetting `&` somewhere), but:
  - we could provide a PHPStan rule to detect at least trivial cases where `&` was forgotten at the callside
  - if the reference is broken, the placeholder will stay in place, and our validation layer could pick these up and report these as developer errors during a migration run
  - we can include a caller trace location in the placeholders to quickly find where the mapping was introduced, this is already implemented right now


mapping-experiment.php
```diff
@@ -0,0 +1,168 @@
+<?php declare(strict_types=1);
+
+// Todo: do not merge this experiment
```
MalteJanz, line 3: Todo: remove this and document the approach properly somewhere else

MalteJanz: I didn't come very far with implementing this as an actual `MappingServiceV2` and using that in our `ProductConverter` to validate this idea further today.

But I'm still curious what others would think about this rather unusual approach 🙂

vintagesucks: Creative! My main concern would be DX (using this right), but this is probably solvable with custom PHPStan rules or extra validation steps. Definitely worth exploring more 🙂

MalteJanz:
I got some initial results for refactoring the SW5 `ProductConverter` (see PR description) and with that will conclude this experiment for now. Feel free to take another look with a real example on how this refactor could look like in practice.

Any feedback is welcome, personally I think we should pick this up in the future again and continue with this refactor.

src/Profile/Shopware/Converter/ProductConverter.php
```diff
@@ -187,29 +189,23 @@ public function convert(
...
     }
 
     /**
      * @param array<string, mixed> $data
      */
     protected function convertMainProduct(array $data): ConvertStruct
     {
-        $containerMapping = $this->mappingService->getOrCreateMapping(
-            $this->connectionId,
+        $containerUuid = &$this->mappingServiceV2->getMapping(
```
jozsefdamokos, line 201: Can we somehow enforce that it's called like this everywhere? This can easily be forgotten/overlooked

src/Profile/Shopware/Converter/ProductConverter.php
```diff
@@ -187,29 +189,23 @@ public function convert(
...
             $data['id'],
-            $this->context
+            true,
         );
-        $containerUuid = $containerMapping['entityId'];
 
         $converted = [];
-        $converted['id'] = $containerUuid;
-        $this->mappingIds[] = $containerMapping['id'];
+        $converted['id'] = &$containerUuid;
```
jozsefdamokos, line 208: Same here, very easy to forget the `&`

src/Migration/Mapping/MappingPromise.php
```diff
@@ -0,0 +1,54 @@
...
+         * if creating, use this Uuid to map to
+         */
+        public ?string $createWith = null,
+    ) {
+    }
+
+    public function resolve(string $uuid): void
+    {
+        $this->reference = $uuid;
+        unset($this->reference);
```
jozsefdamokos, line 52: Why the `unset`?

src/Migration/Mapping/MappingServiceV2.php
```diff
@@ -0,0 +1,320 @@
...
+     * The use of '&' is very important, otherwise it will only copy the placeholder string.
+     * More info on this can be found in the PHP docs:
+     * https://www.php.net/manual/en/language.references.return.php
+     *
+     * @param bool $shouldCreate if the mapping should be created if it doesn't exist yet
+     * @param string|null $createWith if creating, use this Uuid to map to. All the same lookups will resolve to this
+     *
+     * @return string placeholder string which will be resolved later by calling @see MappingServiceV2->resolvePromises()
+     */
+    public function &getMapping(
```
jozsefdamokos, line 55: What if instead of returning a `string` here we change it to return an object instead. Objects are passed by reference by default so we would not need the `&` everywhere. Then we could still resolve them later to use the real uuids. Would that work?

src/Profile/Shopware/Converter/ProductConverter.php
MalteJanz: Yes, for this specific case I guess we could write a PHPStan rule to check that the callsite of this method always contains the `&` 🤔

src/Profile/Shopware/Converter/ProductConverter.php
MalteJanz: Yes and unfortunately this is much harder to detect with static code analysis but fortunately we have the validation layer now, which can either fail or properly report any placeholders it founds (they are not valid UUIDs in the end) and I already added a bit a extra information to them so they tell you exactly where the mapping service was called, which makes it easy again to track down and fix these issues (with the only downside that you have to run the migration and run into this issues to catch them).

https://github.com/shopware/SwagMigrationAssistant/blob/0afb86cedf491a8e5c5eaed5b1ad2005e75ffbd6/src/Migration/Mapping/MappingServiceV2.php#L61-L67

src/Migration/Mapping/MappingPromise.php
MalteJanz: Not strictly necessary here, it just drops the reference and makes `$this-reference` an independent copy again, so further calls to resolve don't change the value again (which might be desired or not, up to discussion I guess 🤷 )

src/Migration/Mapping/MappingServiceV2.php
MalteJanz: The issue then is how would you replace an PHP object with a string in a deeply nested associative array without iterating said array again?

The nice thing about this approach is it lets us change these strings freely after the converter has run on the whole batch 🤔

mitelg: without having fully understood how this thing works, maybe another data structure can help? :thinking: https://www.php.net/manual/en/spl.datastructures.php

MalteJanz:
closed as it would require quite a lot of further effort to finish and it was just an idea to explore. Feel free to pick it up again

PR #146, opened 2026-03-02
fix(snippets): Housekeeping of snippets

marcelbrode:

Some snippets are not correct or should use global default instead

Housekeeping

### 5. Checklist

- [ ] I have created the PR in draft status and only open it when it's ready for review
- [ ] If the change is large: I have thought about splitting it up into smaller PRs
- [ ] I have written tests and if it's a bug fix verified that they fail without my change or that new code is covered by tests
- [ ] I have updated developer-facing release notes if this change affects external developers
- [ ] I have written or adjusted the documentation according to my changes
- [ ] I added the correct package annotation to each `src` file (not strictly necessary for tests)
- [ ] This change has code comments where appropriate, especially for non-obvious lines of code
- [ ] Ensure the pull request title as well as first commit follows conventional commits
- [ ] I have thought about well-defined extension points and marked everything else as `@internal` / `@private`


MalteJanz (approved):
Thanks for your contribution and cleaning up these snippets 👍

ravenokko (approved):

PR #148, opened 2026-03-02
chore: fix typescript types and migration api service

MalteJanz:
Looks like our Typescript setup in our plugin is a little broken right now.

I tried to clean up some wrong types a little bit, but ideally I would also like to:
- remove all `@ts-ignore` comments in `swag-migration.api.service.ts`
- somehow import the `ApiService` type from the core, without importing code into our bundle. So we get proper type information for all the methods / fields of the base class
- make sure TSC / Eslint checks types with the `composer run lint:admin` command as well as in our CI

### DIFF
--- a/src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
+++ b/src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
@@ -78,9 +78,9 @@ export default class MigrationApiService extends ApiService {
         connectionName: string,
         profileName: string,
         gatewayName: string,
-        credentialFields: Record<string, MigrationCredentials>,
+        credentialFields: MigrationCredentials,
         additionalHeaders: AdditionalHeaders = {},
-    ): Promise<ApiResponse<unknown>> {
+    ): Promise<ApiResponse<MigrationEnvironmentInformation>> {
         // @ts-ignore
         const headers = this.getBasicHeaders(additionalHeaders);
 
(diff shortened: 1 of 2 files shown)


src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -56,6 +56,9 @@ export type LogLevelCounts = {
     info: number;
 };
 
+// todo: remove me, this is just to test our typescript linting
+export const aNumber: string = 42;
```
MalteJanz, line 60: @larskemper maybe you can help me investigate what's going on when you are back 🙈 .

I already tried to look at other `tsconfig.json` files of our other plugins but so far didn't find a proper config that only checks our plugin files, but still allows importing type from the core 🤔 

The highlighting in the IDE (PHPStorm) seems to work fine for the most part.

I'm a bit confused right now 🤯

larskemper: yep, exactly my problem doing this in the first place 😅 . Will take a look

MalteJanz: feel free to take over / commit to this PR. Unfortunately I'm in meetings all day, otherwise I would have suggested to take a look together 🙈

larskemper: also not really an idea, but i guess we could do something like this: https://github.com/shopware/SwagMigrationAssistant/pull/149/changes#diff-24da125432db6a98cab560db54b6abf35a172a6023048c271ccce5774cceb6f1R8

at least catches the type issues in our code

larskemper:
covered by https://github.com/shopware/SwagMigrationAssistant/pull/149

PR #149, opened 2026-03-03
fix: ts type check

larskemper:
### Changes

- adds missing typescript type check to ci
- adds missing `axios` dependency for types
- fixes type issues
- refactor api service types
- decoupling of shopware global types


src/Resources/app/administration/package.json
```diff
@@ -15,6 +16,7 @@
   },
   "devDependencies": {
     "@shopware-ag/eslint-config-base": "2.0.0",
+    "axios": "0.30.3",
```
MalteJanz, line 22: I don't think it makes much sense to include our own version of axios as a dependency. I'm thinking if we could somehow use the axios dependency from platform.

For example SwagCommercial has imports like this:
https://github.com/shopware/SwagCommercial/blob/61f80c66a3c17498725ca19adb0af752355c15d2/package.json#L22

Maybe a similiar thing works for us as well🤔

larskemper: In this case, axios is only added as a devDependency and purely for typing purposes. So it's not shipped.
I also checked, but there does not seem to be a separate `@types/axios` package we could use instead.
All other dependencies we define are platform "independent" as well, just pinned to specific versions for consistency. So I would stay consistent with that or change completly

MalteJanz: I'm fine with having it as a dev dependency. I'm only a little concerned because we technically use the axios version from platform. Means if they do a major update of axios, our types might be completely wrong.

So it would be great if you could reuse their exact version for our types 🤷 .

For the other dependencies, yes you are right, we use our own (e.g. ESLint), which is totally fine and also decouples us more from platform (which is a good thing). The only other thing where we technically depend on platform as well would be `Vue` 🙈

larskemper: `0.30.3` is already the version we use for platform ([ref](https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/package.json#L72)). The only thing we would have to do, is with a shopware major to align this

vintagesucks (approved):

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
MalteJanz: I think you forgot to include my changes here if your PR should replace mine 😉 

See things related to the `MigrationCredentials` type (because it was defined wrong as it didn't consider all possible profile + gateway credentials). As well as the return type of this method.

https://github.com/shopware/SwagMigrationAssistant/pull/148/changes

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
```diff
@@ -61,16 +61,10 @@ export type LogLevelCounts = {
...
  */
 export default class MigrationApiService extends ApiService {
-    private readonly basicConfig: AxiosRequestConfig & { version: string };
-
     constructor(httpClient: AxiosInstance, loginService: LoginService, apiEndpoint = 'migration') {
         super(httpClient, loginService, apiEndpoint);
         // @ts-ignore
         this.name = MIGRATION_API_SERVICE;
-        this.basicConfig = {
-            timeout: 30000,
```
MalteJanz, line 71: Why was the timeout removed?

src/Resources/app/administration/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "MIT",
   "scripts": {
     "lint": "eslint --ext .js,.ts,.vue,.html,.html.twig .",
+    "lint:types": "tsc --noEmit 2>&1 | grep -E '^src/' && exit 1 || exit 0",
```
MalteJanz: I mean yes this is already an improvement and will do type checking of our files, but the developer experience isn't that great outside of an IDE. Just comparing the output of this:

Image

With the one `tsc` gives without the extra processing (grep):

Image

So I'm not that satisfied with the solution yet, maybe it's worth asking the admin team if they have another idea on how to structure our typescript project instead of this hacky workaround? 🤔

MalteJanz (commented):
Thanks for taking over, not sure if this is the solution we should accept though or if we can do better than this 🤔

src/Resources/app/administration/src/core/service/api/swag-migration.api.service.ts
larskemper: the default timeout config is already 30s https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/core/factory/http.factory.js#L47

DennisGarding (approved):

src/Resources/app/administration/package.json
larskemper: I explored another possible approach: decoupling the migration from the platform typeScript by not loading the platform global type definitions, which are currently causing the unrelated errors. Most of the required types can instead be imported via packages, with the exception of the global `Shopware` object, which would remain basically untyped. While this might be a cleaner technical solution, it would reduce dx

MalteJanz: I would like such an approach and would actually prefer it over importing files from outside of the plugin folder.

For the `Shopware` object we could also define types, for the little stuff we need, inside our plugin. But would be nice if we could use types e.g. for the `ApiService` or for stuff like the `Component.register` functions from a shopware package 🤔

MalteJanz (approved):

DennisGarding (approved):

PR #150, opened 2026-03-05
fix: exception messages

jozsefdamokos:
- include exception file and line in log entry exception message
- fix exception in ConverterRegistry
- make OrderWriter magento proof
- remove magento 1.9 profile+  add magento 2.4 profile

### DIFF
--- a/src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
+++ b/src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
@@ -40,6 +40,7 @@ class MigrationLogBuilder
         protected ?array $convertedData = null,
         protected ?string $exceptionMessage = null,
         protected ?array $exceptionTrace = null,
+        protected ?\Throwable $exception = null,
     ) {
     }
 
@@ -100,6 +101,13 @@ class MigrationLogBuilder
         return $this;
     }
 
+    public function withException(\Throwable $exception): self
+    {
+        $this->exception = $exception;
+
+        return $this;
+    }
+
     public function withExceptionMessage(string $exceptionMessage): self
     {
         $this->exceptionMessage = $exceptionMessage;
@@ -138,8 +146,8 @@ class MigrationLogBuilder
             $this->fieldSourcePath,
             $this->sourceData,
             $this->convertedData,
-            $this->exceptionMessage,
-            $this->exceptionTrace,
+            $this->getExceptionMessage(),
+            $this->getExceptionTrace(),
         );
     }
 
@@ -151,6 +159,32 @@ class MigrationLogBuilder
 
         if (Uuid::isValid($id)) {
             return $id;
+        }
+
+        return null;
+    }
+
+    private function getExceptionMessage(): ?string
+    {
+        if ($this->exceptionMessage !== null) {
+            return $this->exceptionMessage;
+        }
+
+        if ($this->exception !== null) {
+            return $this->exception->getMessage() . ' in ' . $this->exception->getFile() . ':' . $this->exception->getLine();
+        }
+
+        return null;
+    }
+
+    private function getExceptionTrace(): ?array
+    {
+        if ($this->exceptionTrace !== null) {
+            return $this->exceptionTrace;
+        }
+
+        if ($this->exception !== null) {
+            return $this->exception->getTrace();
         }
 
         return null;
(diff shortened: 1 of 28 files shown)


src/Profile/Shopware/Converter/MainVariantRelationConverter.php
```diff
@@ -35,11 +35,12 @@ public function convert(array $data, Context $context, MigrationContextInterface
         $this->connectionId = $connection->getId();
 
         if (!isset($data['id'], $data['ordernumber'])) {
+            $exception = new \Exception('MainVariantRelation requires ID and order number, to be converted successful');
```
MalteJanz: Would be nice to introduce a `MigrationExeception::mainVariantRelationMissingIdAndOrderNumber` for this instead of a generic one

src/Profile/Shopware/Converter/TranslationConverter.php
```diff
@@ -961,31 +961,30 @@ protected function addAttribute(string $entityName, string $key, string $value,
...
+            $error = $e;
         }
 
-        return $objectData;
+        $this->loggingService->log(
+            MigrationLogBuilder::fromMigrationContext($this->migrationContext)
+                ->withEntityName($entity)
+                ->withFieldSourcePath('objectdata')
+                ->withSourceData($data)
+                ->withException($error)
```
MalteJanz, line 983: Nitpick: I would prefer declaring this `$error` variable above the `try` / `catch` block and initializing it to `null` instead of declaring it in multiple branches

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
```diff
@@ -735,7 +735,7 @@ Profile name: shopware55
 Gateway name: local
 Created at: [timestamp]
 Entity: product
-Exception message: SwagMigrationAssistant\Profile\Shopware\Converter\ProductConverter::getPrice(): Argument #1 ($priceData) must be of type array, null given, called in /home/runner/work/SwagMigrationAssistant/SwagMigrationAssistant/custom/plugins/SwagMigrationAssistant/src/Profile/Shopware/Converter/ProductConverter.php on line 383
+Exception message: SwagMigrationAssistant\Profile\Shopware\Converter\ProductConverter::getPrice(): Argument #1 ($priceData) must be of type array, null given, called in [path]src/Profile/Shopware/Converter/ProductConverter.php on line 383
```
MalteJanz: we should replace the line numbers if not also the file paths here in the snapshot, otherwise the snapshot will fail every time the implementation changes (lines added / removed or file moved). Like we already do for other non deterministic stuff:
https://github.com/shopware/SwagMigrationAssistant/blob/c26f0ade93a0468238967983044b3c313faa2591/tests/acceptance/tests/MigrationTest.spec.ts#L264-L280

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -279,6 +279,9 @@ test.describe('Migration Tests @migration @visual', () => {
                 'Exception trace (JSON):\n[trace]\n',
             );
 
+            // remove exception paths
+            logString = logString.replaceAll(/(^Exception message:.*? in )\/[^ ]*?\/src\//gm, '$1[path]src/');
```
MalteJanz: Nevermind my above comment, but this doesn't seem to be enough / work correctly?

MalteJanz (approved):

src/Profile/Shopware/Converter/TranslationConverter.php
jozsefdamokos: This actually gives a warning because the variable would be overwritten straight away

MalteJanz (approved):

larskemper (approved):

MalteJanz (approved):

src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
```diff
@@ -155,4 +163,33 @@ private function getRevisedId(?string $id): ?string
...
+     * @return array<mixed>|null
+     */
+    private function getExceptionTrace(): ?array
+    {
+        if ($this->exceptionTrace !== null) {
+            return $this->exceptionTrace;
+        }
+
+        if ($this->exception !== null) {
+            return $this->exception->getTrace();
```
MalteJanz: likely needs to be adjusted again to filter out `args` as well, see
https://github.com/shopware/SwagMigrationAssistant/pull/151

MalteJanz (approved):

PR #151, opened 2026-03-09
fix: decrease log size by removing trace args

MalteJanz:
The migration log files can quickly get huge. For example I had a migration run with around 1000 exceptions, which produced a 1.2 GB log file.

With this adjustment a similar run only produces a log file of around 10 MB instead.

I don't think we lose much context by removing the trace args, often they are just the whole converted array to be written, which we already include elsewhere in the log.

### DIFF
--- a/src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
+++ b/src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
@@ -128,6 +128,13 @@ class MigrationLogBuilder
     {
         \assert(\class_exists($logClass) && \is_subclass_of($logClass, AbstractMigrationLogEntry::class));
 
+        if ($this->exceptionTrace !== null) {
+            // remove args from trace, as they quickly become too large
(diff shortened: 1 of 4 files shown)


jozsefdamokos (approved):

DennisGarding (approved):

src/Migration/Logging/Log/Builder/MigrationLogBuilder.php
```diff
@@ -128,6 +128,13 @@ public function build(string $logClass): AbstractMigrationLogEntry
     {
         \assert(\class_exists($logClass) && \is_subclass_of($logClass, AbstractMigrationLogEntry::class));
 
+        if ($this->exceptionTrace !== null) {
```
larskemper, line 131: why not do that in `withExceptionTrace()` directly?

jozsefdamokos: will be moved anyway in https://github.com/shopware/SwagMigrationAssistant/pull/150 😉

PR #153, opened 2026-03-16
refactor: use php service definition

### LINKED ISSUE #15351: Migration-Assistant switch from XML to PHP service definitions
### Technical TODO

- [ ] Migrate all XML definitions to PHP definitions
- [ ] Update dev docs

core context: https://github.com/shopware/shopware/issues/13072

larskemper:
resolves https://github.com/shopware/shopware/issues/15351

### Changes

- converted all xml service definitions to php
- refactored `SwagMigrationAssistant.php`

Completely compatible since symfony `5.1` (shopware has symfony `7.2.0 - 7.4.0`)


src/DependencyInjection/dataProvider.php
```diff
@@ -0,0 +1,410 @@
...
+
+    $services->set(ProviderRegistry::class)
+        ->args([
+            new TaggedIteratorArgument('shopware.dataProvider.provider'),
+        ]);
+
+    $services->set(DataProviderController::class)
+        ->public()
+        ->args([
+            new Reference(ProviderRegistry::class),
```
mitelg: have you used this tool? https://github.com/GromNaN/symfony-config-xml-to-php/

I wonder why you add everything with `new ` and not with the `service()` method 🤔 like Umut did here: https://github.com/shopware/SwagCommercial/pull/2187/changes

src/SwagMigrationAssistant.php
```diff
@@ -104,18 +120,13 @@ public function uninstall(UninstallContext $context): void
...
-DROP TABLE IF EXISTS swag_migration_mapping;
-DROP TABLE IF EXISTS swag_migration_logging;
-DROP TABLE IF EXISTS swag_migration_media_file;
-DROP TABLE IF EXISTS swag_migration_run;
-DROP TABLE IF EXISTS swag_migration_connection;
-');
-
-        parent::uninstall($context);
+
+        $connection->executeStatement('SET foreign_key_checks = 0');
```
mitelg: instead of disabling the FK check, the tables should be deleted in the correct order. this is also something we always reject in reviews for store extensions 😬

src/DependencyInjection/dataProvider.php
larskemper: saw it here https://github.com/shopware/shopware/blob/trunk/src/Core/System/DependencyInjection/consent.php. But will change, looks cleaner with `service` 🤝

src/SwagMigrationAssistant.php
larskemper: yeah, true 🏃

mitelg (approved):

DennisGarding (approved):

PR #154, opened 2026-03-17
fix: fingerprint compatibility

### LINKED ISSUE #15637: Migration assistant v 16.x is not compatible with shopware/core v 6.7.1.0
### Shopware Version

6.7.1.0

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

Attempt to establish migration connection with source system with Shopware 6.7.1.0 and Migration Assistant 16.1.0 ends with an error: `No connection Could not connect to the server. Please check the given shop domain.`

Call to the API endpoint https://source.shopware.instance/api/_action/data-provider/get-environment throws an error: 
```
{
    "errors": [
        {
            "code": "0",
            "status": "500",
            "title": "Internal Server Error",
            "detail": "Undefined constant Shopware\\Core\\Framework\\App\\ShopId\\ShopIdProvider::SHOP_ID_SYSTEM_CONFIG_KEY_V2"
        }
    ]
}
```

referenced constant was introduced in shopware/core 6.7.2.0: https://github.com/shopware/core/blob/21b1fdf6ecde57d8d29b2956e19184a25811e1cc/Framework/App/ShopId/ShopIdProvider.php#L25

however, latest Migration Assistant claims to be compatible with any 6.7.x release: https://github.com/shopware/SwagMigrationAssistant/blob/1ee2c1356e9844563ababe6951212a90f605f9b1/composer.json#L13

### Expected behaviour

Migration assistant contains valid restrictions in composer.json and is only installed on compatible systems

### How to reproduce

Install latest available Migration Assistant (16.1.0 at the moment) on source shop with Shopware 6.7.1.0.
Create integration in source system and configure migration connection in destination instance following instructions from documentation https://docs.shopware.com/en/migration-en/shopware6-Migrationsprocess

larskemper:
resolves https://github.com/shopware/shopware/issues/15637

### Changes

- fall back to v1 shopId if v2 is empty

### DIFF
--- a/src/DataProvider/Service/EnvironmentService.php
+++ b/src/DataProvider/Service/EnvironmentService.php
@@ -9,6 +9,7 @@ namespace SwagMigrationAssistant\DataProvider\Service;
 
 use Shopware\Core\Defaults;
 use Shopware\Core\Framework\Api\Context\SystemSource;
+use Shopware\Core\Framework\App\ShopId\ShopIdProvider;
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
@@ -23,8 +24,11 @@ use Shopware\Core\System\SystemConfig\SystemConfigService;
 #[Package('fundamentals@after-sales')]
 class EnvironmentService implements EnvironmentServiceInterface
 {
-    private const SHOP_ID_CONFIG_KEY_V1 = 'core.app.shopId';
-
+    /**
+     * @see ShopIdProvider::SHOP_ID_SYSTEM_CONFIG_KEY_V2
+     * Which was introduced with 6.7.2.0, but we need to support older versions as
+     * well, so we cannot directly use the constant
+     */
     private const SHOP_ID_CONFIG_KEY_V2 = 'core.app.shopIdV2';
 
     /**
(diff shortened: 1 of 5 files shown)


src/DataProvider/Service/EnvironmentService.php
```diff
@@ -83,7 +83,8 @@ public function getEnvironmentData(Context $context): array
 
     private function getShopIdV2(): ?string
     {
-        $response = $this->systemConfigService->get(ShopIdProvider::SHOP_ID_SYSTEM_CONFIG_KEY_V2);
+        $response = $this->systemConfigService->get(ShopIdProvider::SHOP_ID_SYSTEM_CONFIG_KEY_V2)
```
MalteJanz: Have you checked what happens if you remove that V2 constant from the core class? I guess it might still throw an PHP error that the property doesn't exists 🤔

larskemper: yeah true, have to use the strings

jozsefdamokos (approved):

src/DataProvider/Service/EnvironmentService.php
```diff
@@ -24,6 +23,10 @@
 #[Package('fundamentals@after-sales')]
 class EnvironmentService implements EnvironmentServiceInterface
 {
+    private const SHOP_ID_CONFIG_KEY_V1 = 'core.app.shopId';
+
+    private const SHOP_ID_CONFIG_KEY_V2 = 'core.app.shopIdV2';
```
MalteJanz, line 32: nit: maybe add a comment where these are origintally coming from:

```suggestion
    /**
    * @see Shopware\Core\Framework\App\ShopId\ShopIdProvider::SHOP_ID_SYSTEM_CONFIG_KEY_V2
    * which was only introduced with 6.7.2.0, thats why we couldn't use it directly.
    */
    private const SHOP_ID_CONFIG_KEY_V2 = 'core.app.shopIdV2';
```

MalteJanz (approved):

PR #155, opened 2026-03-18
fix: handle missing prices on product conversion

### LINKED ISSUE #15204: Handle missing prices in ProductConverter and improve editor field error resolution
### Technical TODO

PR shopware/SwagMigrationAssistant#141 uncovered a previously-swallowed exception in `ProductConverter::getPrice()`. The fix attempt (adding a `MigrationValidationRequiredFieldMissingLog` instead) was reverted because it caused ATS failures. The error resolution tests fail when price fields produce a validation error (editor field with `[uuid]` placeholder).

Related discussion: https://github.com/shopware/SwagMigrationAssistant/pull/141#discussion_r2846201744

jozsefdamokos:
fixes https://github.com/shopware/shopware/issues/15204

- added empty checks for prices in ProductConverter
- added early returns in CrossSellingConverter when products dont exist
- added price for product in shopware55 test database
- replace log entry counts with placeholder


tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -287,6 +287,9 @@ test.describe('Migration Tests @migration @visual', () => {
             // (the regex replaces everything after the last occurrence of "called in" in the message)
             logString = logString.replaceAll(/^(Exception message:.*called in ).*$/gm, '$1[path]');
 
+            // replace log entry counts
+            logString = logString.replaceAll(/----- Log Entry #\d+ -----/g, '----- Log Entry [count] -----');
```
larskemper, line 291: instead of replacing them, we could also think about just removing the number from the divider completly: https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Migration/History/HistoryService.php#L128

larskemper (approved):

tests/acceptance/tests/MigrationTest.spec.ts
jozsefdamokos: I think it can still be valuable. Im thinking when we share log files we can say look at entry number 100 for example.

DennisGarding (approved):

PR #156, opened 2026-03-19
test: dynamic entity count ats

### LINKED ISSUE #15245: Adjust migration acceptance test to compare relative entity counts instead of absolute ones
### Technical TODO

See this PR on why the current solution is bad:
https://github.com/shopware/SwagMigrationAssistant/pull/143

We should instead fetch counts once before the migration and once after it, and then compare the relative change with our expected numbers.

larskemper:
resolves https://github.com/shopware/shopware/issues/15245


tests/acceptance/fixtures/EntityCounter.ts
```diff
@@ -1,44 +1,143 @@
...
+    'product_property_relation',
+    'main_variant_relation',
+    'translation',
+]);
+
+/**
+ * The COUNT_CORRECTION object contains the expected difference in entity counts after the
+ * migration compared to the baseline. The difference are caused due to duplicates or invalid/incomplete data
+ */
+const COUNT_CORRECTION = {
```
larskemper: @shopware/product-cc-after-sales  Did not really had a better idea. We could also just check for `toBeGreaterThanOrEqual` / `toBeLessThanOrEqual` instead for these cases

MalteJanz: At first glance, I was quite confused what this `COUNT_CORRECTION` is actually about and the comment didn't help either until I understood that you are computing the expected counts based on the data selection totals that the migration-assistant reports from the source system.

To be honest, I would keep it even simpler and not fetch + compute any "expected counts" from the datasets. Instead as before let the test specify the expected counts (which was totally fine and likely is much less error prone than this computation based on the counts / totals of the source system + correction numbers).

For the test itself, I would suggest a simple API like this:
```ts
const entityCountExpecation = await EntityCounter.buildExpectationWithBaseline({
    'product': 427,
    'category': 63,
    // ...
});
```
And then later call it like so to do the actual assertions:
```ts
await EntityCounter.assertExpecation(entityCountExpecation);
```

You could also keep the line by line asserts like we had before, but then you would need to define all entities for the baseline in advance, which means duplicate code that might not match, that's why I think its better to define the entities with the expected counts once upfront.

jozsefdamokos (approved):

DennisGarding (approved):

vintagesucks (approved):

MalteJanz (approved):
Nicely done 👍

PR #158, opened 2026-03-24
refactor!: centralize configurations

### LINKED ISSUE #14598: make migration assistant configurable
### Technical TODO

We may want to make some configs editable for the user:

- Migration batch size
   Current: 100
   Valid range: 25 - 500 (?)

- Http connection timeout
   Current: 15s
   Valid range: 5 - 120s (?)

- Media error threshold (retry)
   Current: 3
   Valid range: 1 - 10 (?)

- Max exception count (auto abort)
   Current: 3
   Valid range: 1 - 15 (?)

- Media processing batch size
   Current: 10
   Valid range: 1 - 50 (?)

- Data fetch limit
   Current: 250
   Valid range: 50 - 1000 (?)

- Log buffer size (?)
   Current: 50
   Valid range: 10 - 250 (?)

configs that should not be editable:
 
- other batch/limit sizes: mostly pure implementation detail, no real benefit
- temp url expiration: security related
- prefix/paths/directories: could cause conflicts

larskemper:
resolves https://github.com/shopware/shopware/issues/14598

### Changes

- introduced `MigrationConfiguration` service (overridable by injection)


src/Migration/Gateway/HttpSimpleClient.php
```diff
@@ -41,7 +41,7 @@ public function get(string $uri, array $options = []): ResponseInterface
     protected function constructClient(array $additionalOptions = []): Client
     {
         return new Client(\array_merge([
-            'connect_timeout' => 15.0,
+            'timeout' => 15.0,
```
larskemper: ref: https://github.com/guzzle/guzzle/blob/7.10/src/Client.php#L34

DennisGarding (approved):

jozsefdamokos (approved):

src/Migration/MigrationConfiguration.php
```diff
@@ -0,0 +1,31 @@
...
+     * @internal
+     */
+    public function __construct(
+        public int $MIGRATION_LOG_BUFFER_SIZE = 50,
+        public int $MIGRATION_LOG_EXCEPTION_TRACE_ITEM_LIMIT = 10,
+        public int $MIGRATION_DEFAULT_BATCH_SIZE = 250,
+        public int $MIGRATION_MEDIA_PROCESSING_BATCH_SIZE = 10,
+        public int $MIGRATION_DEFAULT_EXCEPTION_THRESHOLD = 3,
+        public int $MIGRATION_DEFAULT_FETCH_SIZE = 50,
+        public int $MIGRATION_CONNECTION_TIMEOUT = 15,
```
vintagesucks: `SCREAMING_SNAKE_CASE` feels a little weird for properties IMO

src/Profile/Shopware/Media/HttpMediaDownloadService.php
```diff
@@ -35,7 +35,7 @@ protected function getMediaEntity(): string
 
     protected function getHttpClient(MigrationContextInterface $migrationContext): ?HttpClientInterface
     {
-        return new HttpSimpleClient();
+        return new HttpSimpleClient(['timeout' => $this->migrationConfig->MIGRATION_CONNECTION_TIMEOUT]);
```
vintagesucks: Do I understand this correctly that there wasn't a timeout before, and there is a 15s one by default now? Could that become a problem for big media files (or slow connections)?

vintagesucks (approved):

src/Migration/Gateway/HttpSimpleClient.php
MalteJanz: There is also `connect_timeout`:
https://docs.guzzlephp.org/en/stable/request-options.html#connect-timeout

Not sure if `timeout` applies the same way but I agree that it makes more sense to use that instead. We might want to do a quick connection check with it though, the worst case could be that you input an wrong domain and try to connect to it and it runs forever (until the browser kills the request).

src/Migration/MigrationConfiguration.php
MalteJanz:
```suggestion
        public int $MIGRATION_REQUEST_TIMEOUT = 15,
```
technically its not only a connection timeout anymore.

src/Profile/Shopware/Media/HttpMediaDownloadService.php
MalteJanz: Previously there was also a 15s timeout but only to establish the HTTP connection, not for the total request lifetime.

I think we should either revert back to only `connect_timeout` or increase this value to something more reasonable (e.g. 30 or 60s)

larskemper: I reverted to old state

src/Migration/MigrationConfiguration.php
vintagesucks: UPGRADE.md still has the old naming

vintagesucks (approved):

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
MalteJanz, line 1: why did this snapshot change?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # NEXT
 
 - #15568 - Switched to PHP Symfony service definitions
+- #14598 - Introduced `MigrationConfiguration` service as a central place for internal configurations
```
MalteJanz: Nit: it's not user facing so no need to mention it here, it's enough to put it in the upgrade file for developers to not confuse any normal users

CHANGELOG_de-DE.md
```diff
@@ -1,6 +1,7 @@
 # NEXT
 
 - #15568 - Umstellung auf PHP-Symfony-Service-Definitionen
+- #14598 - `MigrationConfiguration` als zentrale Stelle für interne Konfigurationen eingeführt
```
MalteJanz: Nit, see above. Same applies to the line above this

MalteJanz (changes requested):
Besides the snapshot file change it looks good to me

CR0YD (approved):

tests/acceptance/snapshots/MigrationTest.spec.ts/linux/migration-log-sw5.txt
larskemper: The default batch size changed from 100 to 250, which changes processing order and so also the order of logs

MalteJanz: That explains it, thanks. But why was it changed? It's not mentioned in the issue or PR description and my assumption of this refactor would be that the logic stays the same unless otherwise explained 😅 

I would vote for keeping it at 100 by default, especially the product entity is quite heavy during read + convert + write and I could imagine multiplying the data by 2,5 can cause some issues on some systems (e.g. memory limits or timeouts).

larskemper: most of the processor defined their batch size themself with `250`, thats why I changed it. But I'm fine with `100`

MalteJanz: Ah I see where this is coming from. But it looks like only cleanup, checksums and truncate processors where using these larger numbers, not the general fetch + convert + write migration flow 😉

MalteJanz (approved):

PR #160, opened 2026-03-25
fix: phpstan

larskemper:
cause of https://github.com/shopware/shopware/pull/15179


tests/integration/Migration/Controller/ErrorResolutionControllerTest.php
```diff
@@ -74,7 +74,7 @@ public function testGetFieldStructureUnknownField(): void
     }
 
     /**
-     * @param array<string, string> $expected
+     * @param array<string, string|false> $expected
```
MalteJanz, line 77: Nit: maybe add an additional assert that `$expected['example']` is not `false`. Because that shouldn't happen in practice and if it would, it means `\json_encode` failed and the data provider produces some strange expecations

MalteJanz (approved):

jozsefdamokos (approved):

PR #161, opened 2026-03-26
chore: sync npm dependencies

larskemper:
### Changes

- synced all npm packages with platform
- cause of major eslint upgrade: migrated to flat config


tests/Jest/jest.config.js
```diff
@@ -73,6 +74,11 @@ module.exports = {
         '^lodash-es/(.*)$': 'lodash/$1',
     },
 
+    transform: {
+        '^.+\\.[jt]sx?$': '<rootDir>/tests/Jest/babel-transform.js',
```
MalteJanz: is this babel-transform really needed? We don't use `jsx` / `tsx` / react 😅

MalteJanz (approved):

tests/Jest/jest.config.js
larskemper: the regex matches ".js, .jsx, .ts, .tsx", so not only react files 😄. The transform itself is definitely needed for jest, but we cloud remove jsx files from the regex

DennisGarding (approved):

vintagesucks (approved):

PR #163, opened 2026-03-30
ci: add nightly

larskemper:
resolves https://github.com/shopware/shopware-private/issues/214

### Changes

- refactored workflows to be reusable
- added caching for npm dependencies
- added concurrency groups (runners with old state get canceled)
- added `nightly.yml` with slack notification
- added `reactivate-workflows.yml` in case github deactivates workflows cause of inactivity


.github/workflows/action-acceptance.yml
```diff
@@ -0,0 +1,98 @@
...
+                ports:
+                    - 3306:3306
+        env:
+            DATABASE_URL: mysql://root:root@127.0.0.1:3306/shopware
+        steps:
+            - name: Cache npm
+              uses: actions/cache@v4
+              with:
+                  path: ~/.npm
+                  key: ${{ runner.os }}-acceptance-${{ hashFiles('tests/acceptance/package-lock.json') }}
```
vintagesucks: I don't think this works without any checkout / extension setup, the path does not exist yet

From https://github.com/shopware/SwagMigrationAssistant/actions/runs/23747827435/job/69181177028:

> Cache hit for: Linux-acceptance-

.github/workflows/action-acceptance.yml
```diff
@@ -0,0 +1,98 @@
...
+
+            - name: Cache Playwright browsers
+              uses: actions/cache@v4
+              with:
+                  path: ~/.cache/ms-playwright
+                  key: playwright-${{ runner.os }}-${{ hashFiles('custom/plugins/**/tests/acceptance/package-lock.json') }}
```
vintagesucks: See above

> Cache hit for: playwright-Linux-

.github/workflows/action-jest.yml
```diff
@@ -0,0 +1,45 @@
...
+        name: Jest
+        runs-on: ubuntu-latest
+        env:
+            ARTIFACTS_PATH: ${{ github.workspace }}/development
+        steps:
+            - name: Cache npm
+              uses: actions/cache@v4
+              with:
+                  path: ~/.npm
+                  key: ${{ runner.os }}-jest-${{ hashFiles('tests/Jest/package-lock.json') }}
```
vintagesucks: See above

> Cache hit for: Linux-jest-

vintagesucks (approved):

.github/workflows/action-acceptance.yml
```diff
@@ -0,0 +1,98 @@
...
+                env:
+                    MYSQL_ROOT_PASSWORD: root
+                    MYSQL_DATABASE: shopware
+                ports:
+                    - 3306:3306
+        env:
+            DATABASE_URL: mysql://root:root@127.0.0.1:3306/shopware
+        steps:
+            - name: Cache npm
+              uses: actions/cache@v4
```
vintagesucks:
```suggestion
              uses: actions/cache@v5
```

but do we also want to pin all our action versions to sha here?

.github/workflows/action-acceptance.yml
larskemper: good one, wrong order & path 😩. should be fixed

.github/workflows/action-acceptance.yml
larskemper: i updated to v5

> but do we also want to pin all our action versions to sha here?

What is the benefit 😅 ?

vintagesucks:
```yml
uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
```

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

> Pinning an action to a full-length commit SHA is currently the only way to use an action as an immutable release.

`@v5` is just a tag: https://github.com/actions/cache/releases/tag/v5

vintagesucks: TIL: There even is a [Require actions to be pinned to a full-length commit SHA](https://github.com/shopware/SwagMigrationAssistant/settings/actions) setting, but I have no idea how that would behave in combination with our reusable workflows.

* https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning

larskemper: good to know 🤝
I pinned them to 5.0.4

larskemper: we could try the setting and see what happens 😄

DennisGarding (approved):

.github/workflows/action-acceptance.yml
larskemper: tried it and it does fail, cause https://github.com/shopware/github-actions are not pinned

jozsefdamokos (approved):

.github/workflows/action-acceptance.yml
MalteJanz: Good to know, thanks for trying. Would be great if you could raise an issue for that repo to pin all its actions (or raise an issue in shopware for the product ops team) 🙂

PR #164, opened 2026-03-30
fix: phpstan

larskemper:


DennisGarding (approved):

tests/Core/Migration/Migration1754896654TruncateMigrationLogsTest.php
```diff
@@ -20,7 +19,6 @@
  * @internal
  */
 #[Package('fundamentals@after-sales')]
-#[CoversClass(Migration1754896654TruncateMigrationLogs::class)]
```
jozsefdamokos, line 23: Why remove this? Looking at the CoversAttributeRule migrations tests should have it, right?

tests/unit/Migration/Logging/Log/MigrationLogTest.php
```diff
@@ -56,7 +54,6 @@
  * @internal
  */
 #[Package('fundamentals@after-sales')]
-#[CoversClass(MigrationLogEntry::class)]
```
jozsefdamokos, line 59: Also this and many others are unit tests so `CoversClass` makes sense

jozsefdamokos (changes requested):

tests/Core/Migration/Migration1754896654TruncateMigrationLogsTest.php
larskemper: for sure false positive, will ignore then

jozsefdamokos (approved):

PR #165, opened 2026-04-02
fix: MediaProcessingProcessor

DennisGarding:
closes: #https://github.com/shopware/shopware/issues/15931

Media processing no longer gets stuck on invalid entries or step over during selection and only fetches media files that still need processing


jozsefdamokos (approved):

CR0YD (approved):

src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
```diff
@@ -179,8 +200,9 @@ private function getMediaFiles(MigrationContextInterface $migrationContext): arr
             ->from('swag_migration_media_file')
             ->where('run_id = :runId')
             ->andWhere('written = 1')
-            ->orderBy('entity, file_size')
-            ->setFirstResult($migrationContext->getOffset())
+            ->andWhere('processed = 0')
+            ->andWhere('process_failure = 0')
+            ->orderBy('id, file_size, entity')
```
larskemper, line 205: isn't order by uuid effectively meaningless?


```suggestion
            ->orderBy('file_size, entity')
```

larskemper (approved):

src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
DennisGarding: At this point file_size and entity could all be the same. `FileSize = 0` and `entity = order_document`
To have a lined-up order, the ID is the best way

PR #166, opened 2026-04-07
ci: add migration proof workflow

### LINKED ISSUE #10824: Evaluate the possibilty of CI migration tests
### Technical TODO

It would be great to have a fully end to end tested migration run in our CI pipeline. This would give the following benefits:

Detect major breaks in platform, that causes migrations to miss data

Gives us more confidence that our code changes work

Could potentially test all our supported profiles in parallel, something that is done almost never locally (because you need all the source systems installed locally to test)

This evaluation ticket should research the following questions for our CI:

Is it possible to run two shopware instances in parallel?

If not possible in GitLab CI, also check GitHub actions, we could switch theoretically

can they communicate with each other?

Is it possible to setup a source system (with demodata, plugin installation, API integration)?

Is it possible to setup the target system (clean, with current plugin code) and run a migration?

Either e2e tested via the brower

Or CLI or even better both

Is it possible to run some final test code, which can access both databases?

verify that the common entity counts are the same (for example same amount of products, …)

check the swag_migration_logging table in the target system, which should only contain known / non critical rows

Some helpful resources:

https://docs.gitlab.com/ee/ci/services/ 

It would be great to have a prove of concept at the end of this ticket which runs a simple migration in the pipeline. Also consider follow up tickets for multiple source systems and migration profiles

larskemper:
resolves https://github.com/shopware/shopware/issues/10824

### Changes
- added new test package `tests/migration-proof`
- added `compatibility.yaml` workflow for checking e2e migrations with all available migration profiles

=> This is meant to be a poc, so feel free to share you opinion on the general approach


MalteJanz (changes requested):
Feedback on this PoC, only had a very rough look:

What I like:
- CI matrix for all SW5 profiles
- Running the source systems in Docker containers

What I don't like (yet):
- Lots of code we would need to maintain
- Why reinvent the wheel and build a custom test suite instead of reusing our existing acceptance tests using playright?
  - Can't we make our acceptance tests more generic, e.g. via Env variables configure what profile they select and also reflect that in snapshot names?
  - If making them generic is not an option, we could also write different tests that are only executed in certain environments and abstract most of the test code away
- I don't understand this `reduce-dabase-for-proof.sql` thing at first glance, why is that needed (looks like it truncates most of the SW5 database)?
- I'm not a fan of `entity-counts.json`, especially the sourceQuery. Why is that needed to compare relative entity count changes in the target system after the migration?
- In general the approach looks rather hard coded for SW5 with DB gateway migrations, what about testing the API gateway? SW6? Magento?

What I would suggest:
- Start from scratch
- Look into ways to make our existing acceptance test generic enough to perform migrations for any profile + gateway combination
- Look into ways to run all our possible source systems in Docker containers (I guess you can reuse some of the stuff here for SW5), with different post processing (e.g. SW5 Connector plugin install, demodata)
- Setup the CI job to just run our acceptance test against different source systems with different profile + gateway combinations and make use of snapshot tests to persist and commit the results once

If you use AI for these suggestions, which I would encourage to try, please make sure to give it strong guardrails and try to keep the solution simple and minimal. To be honest I don't want to fix a bug in this proposed testsuite and I think we can do a lot better here 🙂

larskemper:
@MalteJanz Thanks for the feedback, this is helpful.

A few short answers on why I went this way:

I did not start from the existing acceptance tests because they are currently tied quite closely to one specific setup and fixture, and I did not want to disrupt that too much for a poc.

Even if we used Playwright, we would still need most of the same orchestration code: provisioning both systems, installing plugins, waiting for readiness, preparing the source, starting the migration environment, and collecting artifacts. From my perspective, the main difference would be replacing api calls with browser interactions, not actually removing much of the complexity.

That is also why I did not choose Playwright for the poc. For this kind of infrastructure heavy migration proof, using an e2e browser framework to orchestrate the whole process felt like a stretch and like an unnatural use of Playwright. It seemed more natural to validate the setup first through api driven execution.

`reduce-database-for-proof.sql` was only meant to keep ci runtime and fixture size manageable. I agree that it is too implicit right now and would need better explanation or a cleaner replacement. I also wanted to avoid adding another full database dump to the codebase 😅

`entity-counts.json` was meant as a lightweight regression check so we do not rely only on screenshots or manual log inspection. I agree that, in its current form, it is too custom and too hardcoded.

So I do agree that parts of the poc should be simplified or reworked. I am just not fully convinced that moving the whole flow into Playwright would automatically make it better or cleaner, because the main complexity here is the environment orchestration itself, not whether the migration is triggered through api calls or browser interactions.

MalteJanz:
@larskemper Thanks for the insights. The "why I went this way" explanation would have been great if it were in the PR description from the start 😉 .

> Even if we used Playwright, we would still need most of the same orchestration code: provisioning both systems, installing plugins, waiting for readiness, preparing the source, starting the migration environment, and collecting artifacts. From my perspective, the main difference would be replacing api calls with browser interactions, not actually removing much of the complexity.
>
> That is also why I did not choose Playwright for the poc. For this kind of infrastructure heavy migration proof, using an e2e browser framework to orchestrate the whole process felt like a stretch and like an unnatural use of Playwright. It seemed more natural to validate the setup first through api driven execution.
>
> ... I am just not fully convinced that moving the whole flow into Playwright would automatically make it better or cleaner, because the main complexity here is the environment orchestration itself, not whether the migration is triggered through api calls or browser interactions.

Yes, and I also wouldn't say we should use Playwright to do all this orchestration. What I would like is that we do all this orchestration elsewhere, and then Playright can just be used as the actual acceptance test, running the migration in the target system and validating snapshots. It shouldn't care or interact with the source system at all.


> reduce-database-for-proof.sql was only meant to keep ci runtime and fixture size manageable. I agree that it is too implicit right now and would need better explanation or a cleaner replacement. I also wanted to avoid adding another full database dump to the codebase 😅

One other idea that came to my mind: What if we would create a new repository, that:
- Stores orchestration, preparation, SQL dumps, ... for each source system we want to test
- Maybe it can build custom docker containers, where all the preparation is already included
- And then in our main repo here we just use these docker containers in CI, wait that their up and healthy, and then run Playright tests to do the migration 🤔
- This comes with a slight downside that we have another repository, but I don't think we need to do changes to these source systems often and they should stay rather stable (basically serve as demo data instances)

larskemper:
>The "why I went this way" explanation would have been great if it were in the PR description from the start 😉 .

true 😅 
 
> One other idea that came to my mind: What if we would create a new repository, that:
> 
> * Stores orchestration, preparation, SQL dumps, ... for each source system we want to test
> * Maybe it can build custom docker containers, where all the preparation is already included
> * And then in our main repo here we just use these docker containers in CI, wait that their up and healthy, and then run Playright tests to do the migration 🤔
> * This comes with a slight downside that we have another repository, but I don't think we need to do changes to these source systems often and they should stay rather stable (basically serve as demo data instances)

Actually, that was my first idea as well. I did not go in that direction initially because I wanted to avoid turning this into a new "product" of its own 😄

I think the split itself makes a lot of sense though:

- orchestration and source-system preparation outside of the acceptance test
- Playwright only as the acceptance layer for running the migration in the target system and validating snapshots

That feels much cleaner than forcing Playwright to deal with the full environment setup.

My only concern with a separate repository would be that it is definitely not less complex, and we could end up building something that has the same kind of maintenance burden as the ats suite in the long run. So I would want to be careful not to over engineer it too early 😅

But conceptually I do like the idea a lot. I will try to explore that direction next.

larskemper:
fyi: i will continue here https://github.com/shopware/shopware-migration-test-suite

PR #170, opened 2026-04-09
fix!: acl api roles

larskemper:
related https://github.com/shopware/shopware-private/issues/217

### Changes

- `/api/_action/migration/check-connection` should have `swag_migration.editor` as acl role ([finding](https://chatgpt.com/codex/security/findings/8fe2379f502c8191b25fa4fcc1b3bd24))
- `/api/_action/migration/download-logs-of-run` should require authentication `auth_required = true` ([finding](https://chatgpt.com/codex/security/findings/60a240fe9f908191a715c1c5ee67b1fc))

### DIFF
--- a/src/Controller/HistoryController.php
+++ b/src/Controller/HistoryController.php
@@ -71,7 +71,7 @@ class HistoryController extends AbstractController
         path: '/api/_action/migration/download-logs-of-run',
         name: 'api.admin.migration.download-logs-of-run',
         methods: [Request::METHOD_POST],
-        defaults: ['auth_required' => false, PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']]
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']]
     )]
     public function downloadLogsOfRun(Request $request, Context $context): StreamedResponse
     {
--- a/src/Controller/StatusController.php
+++ b/src/Controller/StatusController.php
@@ -283,7 +283,7 @@ class StatusController extends AbstractController
     #[Route(
         path: '/api/_action/migration/check-connection',
         name: 'api.admin.migration.check-connection',
-        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']],
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.editor']],
         methods: [Request::METHOD_POST]
     )]
     public function checkConnection(Request $request, Context $context): JsonResponse
--- a/src/Resources/app/administration/src/module/swag-migration/page/swag-migration-history-detail-errors/index.ts
+++ b/src/Resources/app/administration/src/module/swag-migration/page/swag-migration-history-detail-errors/index.ts
@@ -16,7 +16,6 @@ export interface SwagMigrationHistoryDetailErrorsData {
     sortDirection: string;
     disableRouteParams: boolean;
     limit: number;
-    downloadUrl: string;
 }
 
 /**
@@ -32,6 +31,7 @@ export default Shopware.Component.wrapComponentConfig({
 
     mixins: [
         Mixin.getByName('listing'),
+        Mixin.getByName('notification'),
     ],
 
     props: {
@@ -50,7 +50,6 @@ export default Shopware.Component.wrapComponentConfig({
             sortDirection: 'DESC',
             disableRouteParams: true,
             limit: 10,
-            downloadUrl: '',
         };
     },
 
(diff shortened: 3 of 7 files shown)


DennisGarding (approved):

src/Controller/HistoryController.php
```diff
@@ -71,7 +71,7 @@ public function getGroupedLogsOfRun(Request $request, Context $context): JsonRes
         path: '/api/_action/migration/download-logs-of-run',
         name: 'api.admin.migration.download-logs-of-run',
         methods: [Request::METHOD_POST],
-        defaults: ['auth_required' => false, PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']]
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']]
```
MalteJanz, line 74: Have you manually tested and confirmed that the download still works fine? I could imagine that the `auth_required` flag was there on purpose to support the way JS was calling it 🤔

src/Controller/StatusController.php
```diff
@@ -283,7 +283,7 @@ public function createNewConnection(Request $request, Context $context): JsonRes
     #[Route(
         path: '/api/_action/migration/check-connection',
         name: 'api.admin.migration.check-connection',
-        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.viewer']],
+        defaults: [PlatformRequest::ATTRIBUTE_ACL => ['swag_migration.editor']],
```
MalteJanz, line 286: Is `editor` really the right call here? Because I think every time you open the migration module we perform a connection check and if you are only `viewer`, what else is left that you could view?

src/Resources/app/administration/src/module/swag-migration/page/swag-migration-history-detail-errors/index.ts
```diff
@@ -156,8 +154,27 @@ export default Shopware.Component.wrapComponentConfig({
             return 'swag-migration.index.error-resolution.codes.unknown';
         },
 
-        submitDownload() {
-            this.$refs.downloadForm.submit();
+        async submitDownload() {
+            try {
+                const blob = await this.migrationApiService.downloadLogsOfRun(this.migrationRun.id);
```
MalteJanz, line 159: Not sure if this is best practice, would need to research that myself. I could imagine that this could be annoying for larger files (which the log files could be). Imagine you clicking on download and nothing happens for 20s because you have a slow connection and its a bigger file. Then suddenly you get your browser download showing up and finishing instantly.

I'm not sure how this behaves but it's worth a closer look. I think the old implementation let the browser do the download from an URL, which would show proper progress in the download UI of the browser 🤔

src/Controller/HistoryController.php
larskemper: That was the reason why we had it before. I changed the frontend access away from the unauthenticated form submit & of course verified that its working (the ats btw. includes this flow also)

src/Controller/StatusController.php
larskemper: The endpoint updates the connection actually, but your right with module init. Will revert

src/Resources/app/administration/src/module/swag-migration/page/swag-migration-history-detail-errors/index.ts
larskemper: Agree on the ux concern. Still weird to keep it unauthenticated, since the logs contain quite some information. how are we doing that with import / export? do we have something like a download token or so?

MalteJanz: I think we use multiple approaches in shopware, e.g.:
- Storefront is authenticated with a cookie most of the time, and also uses deeplink codes, and then let the browser do a normal download via URL I think (didn't check the frontend code)
- normal Media files in the admin are basically unprotected, because they are also used in the storefront. So if you download a file there its also accessible via URL without extra authentication, e.g. you can copy the download link and paste it in a private window and it shows the image (But again I didn't check the frontend code)
- documents (private media files) in the admin require authentication for download (bearer token in the header), and I think they are downloaded via JS first like you changed it to here (also didn't check in detail). But most of the time these aren't particularly huge.
- I think import / export does the same, it also requires authentication. That files on the other hand can also become quite large
- And same would apply to our migration logs here

The question is: does this really affect UX? I couldn't test it so far, I tried exporting and that produced a 2MB file but when I turn on network throttling in my browser, it still downloads instantly (likely because it isn't throttled). So maybe this is totally fine, or Import / Export uses yet a different approach than you here 🤷

larskemper: alright, will look into that

PR #171, opened 2026-04-09
fix: skip malformed relation records with missing root id

DennisGarding:
These fixes results from the support ticket #SWAG-327644 & #SWAG-328077

Fix 1: The cover image in products is not set. 
Fix 2: Null or invalid FIX IDs break the migration queue.

### DIFF
--- a/src/Migration/ErrorResolution/MigrationErrorResolutionService.php
+++ b/src/Migration/ErrorResolution/MigrationErrorResolutionService.php
@@ -83,7 +83,14 @@ readonly class MigrationErrorResolutionService
      */
     private function loadFixes(MigrationErrorResolutionContext $errorResolutionContext): void
     {
-        $itemIds = \array_column($errorResolutionContext->getData(), 'id');
+        $itemIds = \array_values(
+            array_filter(
+                \array_column($errorResolutionContext->getData(), 'id'),
+                static function ($itemId): bool {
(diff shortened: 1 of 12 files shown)


src/Migration/ErrorResolution/MigrationErrorResolutionService.php
```diff
@@ -83,7 +83,14 @@ public function applyFixes(array &$data, string $connectionId, string $runId, Co
      */
     private function loadFixes(MigrationErrorResolutionContext $errorResolutionContext): void
     {
-        $itemIds = \array_column($errorResolutionContext->getData(), 'id');
+        $itemIds = \array_values(
```
larskemper, line 90: but how / where are these invalid ids created / written? We probably want to prevent this instead of filtering them.

DennisGarding: Result of talk: is Unknown where the issue cames from

larskemper (approved):

src/Migration/ErrorResolution/MigrationErrorResolutionService.php
larskemper: after investigation: some converters did not check if 'entityId' of read data is empty and created an invalid migration data record that later caused issues with error resolution / writting / cleanup

CR0YD (approved):

PR #177, opened 2026-04-13
fix: phpstan empty not allowed

### LINKED ISSUE #16031: Enforce `empty.notAllowed` in migration assistant
### Technical TODO

Platform introduced phpstan rule `empty.notAllowed`, that we currently ignore.  We should implement more strict comparison and remove the ignore then

larskemper:
resolves https://github.com/shopware/shopware/issues/16031

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -56,9 +56,6 @@ parameters:
         - identifier: parameterByRef.type
           path: src/Profile/Shopware/Converter/ShippingMethodConverter.php
 
-        # TODO: https://github.com/shopware/shopware/issues/16031
-        - identifier: empty.notAllowed
-
 rules:
     - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\Deprecation\DeprecatedMethodsThrowDeprecationRule
     - Shopware\Core\DevOps\StaticAnalyze\PHPStan\Rules\Tests\CoversAttributeRule
--- a/src/Controller/HistoryController.php
+++ b/src/Controller/HistoryController.php
@@ -108,13 +108,13 @@ class HistoryController extends AbstractController
     {
         $runId = $request->query->getAlnum('runId');
 
-        if (empty($runId)) {
+        if ($runId === '') {
             throw RoutingException::missingRequestParameter('runId');
         }
 
         $level = $request->query->getAlpha('level');
 
-        if (empty($level)) {
+        if ($level === '') {
             throw RoutingException::missingRequestParameter('level');
         }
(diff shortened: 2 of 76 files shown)


larskemper:
I couldn't find a concrete reason for the changed log output. The migrated entity counts are unchanged, except for `swag_migration_logging`, which now contains fewer entries. As the migration result itself seems unchanged, I lean toward updating the snapshot.

ref: https://github.com/shopware/SwagMigrationAssistant/actions/runs/24341839999/job/71072462684?pr=177

MalteJanz:
@larskemper 

> I couldn't find a concrete reason for the changed log output. The migrated entity counts are unchanged, except for swag_migration_logging, which now contains fewer entries. As the migration result itself seems unchanged, I lean toward updating the snapshot.

I am pretty sure there is a change in this PR that caused the test failure and would rather not just update the snapshot. Looking at the log file artifact from the CI run:
- Our snapshots expects lots of logs about media files it couldn't "download", which is correct as we don't provide the proper setup for the local gateway to copy these files from a real SW5 instance. In sum we expect 600 `SWAG_MIGRATION_MEDIA_FILE_MISSING` log entries
- But the test run only has 555 of these log entries, means 45 are missing
- And I don't think we suddenly successfully migrated those 45 media files, like how should that be possible in our current CI acceptance test setup without a real SW5 instance?
- Means some change in this PR cause these media items to be skipped without being reported as warning logs

So I think you have to take a deeper look:
- Unfortunately I can't see any obvious pattern on those missing media log entries
- I would start debugging by looking at the numbers of entries in the `swag_migration_media` table after a migration on this branch and trunk, comparing if they are similiar with the same SW5 dataset
- If they are, the issue is likely somewhere in the media processors
- If they don't the issue is likely earlier, either in some converter (like the product converter that creates media entries) or even earlier at the reader stage

larskemper:
@MalteJanz and suddenly after rebasing its green again 😭

MalteJanz:
> [Malte Janz (@MalteJanz)](https://github.com/MalteJanz) and suddenly after rebasing its green again 😭

oh no, lets hope its not flacky (maybe trigger it a few more times ) 🤞 😅. But at least you didn't need to debug further 🙂

larskemper:
> oh no, lets hope its not flacky (maybe trigger it a few more times ) 🤞 😅. But at least you didn't need to debug further 🙂

lets it run a few times, no further fails 🙏

DennisGarding (approved):

jozsefdamokos (approved):

PR #178, opened 2026-04-13
fix!: unsupported sales channel migration

### LINKED ISSUE #14435: Migration of social shopping sales channels fails
### Shopware Version

6.7 / cloud

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

Right now migrations with the Migration-Assistant from SW6 to SW6 seem to fail if the source system contains social shopping sales channels.
Likely because these types of sales channels aren't migrated by the Assistant yet, and then all entities that reference them (like products that are visible in such a sales channel) fail to migrate.

### Expected behaviour

We should look into properly migrating these sales channels or in case that isn't wanted / the target system can't have the social shopping extension, look into if we can drop all associations to these sales channels, so at least entities like products are correctly migrated.

### How to reproduce

Have a SW6.7 instance with the social shopping extension (e.g. SaaS). Create "social shopping" sales channels and assign products to them. Then migrate into another SW6.7 instance.

Hint: the sales channels are likely filtered out here (on the source system side's provider):
https://github.com/shopware/SwagMigrationAssistant/blob/5a9afd81133508891430770194d4ee94d41ce6cc/src/DataProvider/Provider/Data/SalesChannelProvider.php#L46-L49

DennisGarding:
closes:  https://github.com/shopware/shopware/issues/14435

Improve migration of sales-channel for Shopware 6


src/Migration/Mapping/Lookup/SalesChannelLookup.php
```diff
@@ -0,0 +1,60 @@
...
+        $criteria->addFilter(new MultiFilter(
+            MultiFilter::CONNECTION_AND,
+            [
+                new EqualsFilter('typeId', $typeId),
+                new EqualsFilter('translations.name', $name),
+            ]
+        ));
+        $criteria->setLimit(1);
+
+        return $this->cache[$cacheKey] = $this->salesChannelRepository->searchIds($criteria, $context)->firstId() !== null;
```
larskemper: nit: readability (same for the other lookup)


```suggestion
        $this->cache[$cacheKey] = $this->salesChannelRepository->searchIds($criteria, $context)->firstId();

        return $this->cache[$cacheKey] !== null;
```

src/Migration/Mapping/Lookup/SalesChannelLookup.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @param EntityRepository<SalesChannelCollection> $salesChannelRepository
+     */
+    public function __construct(
+        private readonly EntityRepository $salesChannelRepository,
+    ) {
+    }
+
+    public function hasSalesChannelWithTypeAndName(string $typeId, string $name, Context $context): bool
```
larskemper: the other lookup classes implement something like this just with a `get(): ?string` and leave it up to the caller to evaluate, do we want to apply this here also?

src/Profile/Shopware6/Converter/CustomerWishlistConverter.php
```diff
@@ -40,9 +40,11 @@ protected function convertData(array $data): ConvertStruct
 
         $salesChannelMapping = $this->getMappingIdFacade(DefaultEntities::SALES_CHANNEL, $converted['salesChannelId']);
         if ($salesChannelMapping === null) {
-            $converted['salesChannelId'] = null;
+            return new ConvertStruct(null, $data);
```
larskemper, line 54: do we want to log this? `ConvertAssociationMissingLog`? or make it available for the error resolution by not early return (not in this case, when you fetch the mapping, but for others when accessing `$converted['salesChannelId']`)

src/DataProvider/Provider/Data/SalesChannelDomainProvider.php
```diff
@@ -7,11 +7,9 @@
```
larskemper, line 7: please update the changelogs and document breaking changes

src/Migration/Mapping/Lookup/SalesChannelLookup.php
DennisGarding: i think its ok because we also have functions in lookups like `getByNames`  and  `getIsLocked` 

i can rename it to `getHasSalesChannelTypeAndName` but is it necessary?

src/Profile/Shopware6/Converter/CustomerWishlistConverter.php
DennisGarding: i think we wont have this in the error mapping. And before it is also silently failed. because we didnt fetch these sales channels because of the filter.

src/DataProvider/Provider/Data/SalesChannelDomainProvider.php
DennisGarding: Is it a breaking change?

src/Migration/Mapping/Lookup/SalesChannelLookup.php
larskemper: what i meant was to not just return a bool, but the ids of matching sales channels `get` instead of `has` => `getHasSalesChannelTypeAndName(): ?string[]` like the other lookups

src/Profile/Shopware6/Converter/CustomerWishlistConverter.php
larskemper: fine for me in the other cases, but in this specific one, I would log it like the other early returns:

```php
$this->loggingService->log(
                MigrationLogBuilder::fromMigrationContext($migrationContext)
                    ->withEntityName(CustomerWishlistDefinition::ENTITY_NAME)
                    ->withFieldName('salesChannelId')
                    ->withSourceData($data)
                    ->build(ConvertAssociationMissingLog::class)
            );
```

src/DataProvider/Provider/Data/SalesChannelDomainProvider.php
larskemper: `SalesChannelConverter` has a constructor change, right?

UPGRADE.md
```diff
@@ -1,3 +1,10 @@
+# 16.2.0
```
larskemper: `16.2.0` is already released, just use `NEXT`

src/Migration/Mapping/Lookup/SalesChannelLookup.php
DennisGarding: ah, ok, i can switch it

UPGRADE.md
DennisGarding: yes, currently move it to #NEXT

larskemper (approved):

src/Migration/Mapping/Lookup/SalesChannelTypeLookup.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        if (\array_key_exists($salesChannelTypeId, $this->cache)) {
+            return $this->cache[$salesChannelTypeId];
+        }
+
+        $criteria = new Criteria([$salesChannelTypeId]);
+        $criteria->setLimit(1);
+
+        return $this->cache[$salesChannelTypeId] = $this->salesChannelTypeRepository->searchIds($criteria, $context)->firstId();
+    }
```
MalteJanz: At first glance I was really confused by this: Why would you pass in a UUID of a sales channel type and get back the same UUID?

After taking a closer look, it makes a little more sense: the sales channel type entities always have a hard coded UUID, e.g.
https://github.com/shopware/shopware/blob/9ccd0b679f4af93c1ef2dd5ee7b179fb3ab3b6da/src/Core/Defaults.php#L27-L33
and
https://github.com/shopware/SwagSocialShopping/blob/ea7ffd8e166680515ee8dadd6c9513922b232f98/src/SwagSocialShopping.php#L24

This is not something usual and normally we can't rely on any UUID from the source system to make a lookup in the target system (and would use some other unique identifier such as a technical name, which doesn't exist here).

But I would still suggest:
- documenting this fact as a code comment above this method
- I would prefer renaming and changing its signature, so its also totally clear that this is just a check for existence of a type in the shop and returning a boolean

src/Profile/Shopware6/Converter/SalesChannelConverter.php
```diff
@@ -29,16 +45,33 @@ protected function convertData(array $data): ConvertStruct
...
+        $shouldAppendMigrationSuffix = false;
+
         if ($converted['id'] === TestDefaults::SALES_CHANNEL) {
             $mapping = $this->getMappingIdFacade(DefaultEntities::SALES_CHANNEL, $data['id']);
             $converted['id'] = $mapping ?? Uuid::randomHex();
-            $converted['name'] .= ' (Migration)';
+            $shouldAppendMigrationSuffix = true;
+        } elseif (
+            $converted['typeId'] === Defaults::SALES_CHANNEL_TYPE_STOREFRONT
+            && $this->salesChannelLookup->getSalesChannelWithTypeAndName($converted['typeId'], $converted['name'], $this->context) !== null
```
MalteJanz: I don't fully understand this yet. What edge case does this extra if branch cover and is it worth the extra lookup cost? As far as I'm aware the sales channels only have to have a unique primary key `id` and unique `access_key`. Other than that we can just migrate them with their original name. 🤔 

The only edge case is the default sales channel above, because it's `id` is not unique, thus we change it and also rename it to make it clear that that is the migrated one. That's at least my current understanding

src/Profile/Shopware6/Converter/NewsletterRecipientConverter.php
```diff
@@ -37,6 +37,10 @@ protected function convertData(array $data): ConvertStruct
         $converted['languageId'] = $this->getMappingIdFacade(DefaultEntities::LANGUAGE, $data['languageId']);
         $converted['salesChannelId'] = $this->getMappingIdFacade(DefaultEntities::SALES_CHANNEL, $data['salesChannelId']);
 
+        if ($converted['salesChannelId'] === null) {
+            return new ConvertStruct(null, $data);
```
MalteJanz, line 51: As Lars already pointed out above, we should not silently do nothing. Yes in this case, `newletter_recipient` needs a `salesChannelId`, otherwise it can't be inserted in the DB.

And under the assumption that this is likely missing because we didn't migrate that particular sales channel (which also should already created a log), I would say its fine to skip it and I don't see much point in letting users fix it by manually assigning another sales channel.

But the users should still be aware that X newsletter recipients weren't migrated (even though its a follow up issue). So we should definitely create a log entry for these cases.

src/Profile/Shopware6/Converter/OrderConverter.php
```diff
@@ -59,6 +59,10 @@ protected function convertData(array $data): ConvertStruct
             $converted['salesChannelId']
         );
 
+        if ($converted['salesChannelId'] === null) {
+            return new ConvertStruct(null, $data);
```
MalteJanz, line 73: see above

src/Profile/Shopware6/Converter/PageSystemConfigConverter.php
```diff
@@ -38,7 +38,19 @@ protected function convertData(array $data): ConvertStruct
...
-        $systemConfigUuid = $this->systemConfigLookup->get($data['configurationKey'], $data['salesChannelId'] ?? null, $this->context);
+        $mappedSalesChannelId = null;
+        if (isset($data['salesChannelId'])) {
+            $mappedSalesChannelId = $this->getMappingIdFacade(
+                DefaultEntities::SALES_CHANNEL,
+                $data['salesChannelId']
+            );
+
+            if ($mappedSalesChannelId === null) {
+                return new ConvertStruct(null, $data);
```
MalteJanz, line 59: In this case `salesChannelId` is nullable in the `system_config` DB table. Means in theory we could migrate these entries, as long as their `salesChannelId` + `configurationKey` combination is unique.

But in practice I think it makes more sense to skip these as well, because you don't want to mess up your (global) system config that was only configured for a particular sales channel in the source system.

TLDR: keep the early return, but add a log about it as well

src/Profile/Shopware6/Converter/ProductConverter.php
```diff
@@ -201,13 +201,31 @@ protected function convertData(array $data): ConvertStruct
...
+                    continue;
+                }
+
+                $salesChannelId = $this->getMappingIdFacade(
+                    DefaultEntities::SALES_CHANNEL,
+                    $visibility['salesChannelId']
+                );
+
+                if ($salesChannelId === null) {
+                    continue;
```
MalteJanz, line 217: I guess this might be the only place where it's totally fine to silently ignore the data related to the non existing sales channel and don't even log it, because there isn't really any data lost / missing, other than an association to a non existing sales channel (which already produced logs).

src/Profile/Shopware6/Converter/ProductReviewConverter.php
```diff
@@ -38,6 +38,10 @@ protected function convertData(array $data): ConvertStruct
         $converted['salesChannelId'] = $this->getMappingIdFacade(DefaultEntities::SALES_CHANNEL, $data['salesChannelId']);
         $converted['languageId'] = $this->getMappingIdFacade(DefaultEntities::LANGUAGE, $data['languageId']);
 
+        if ($converted['salesChannelId'] === null) {
+            return new ConvertStruct(null, $data);
```
MalteJanz, line 52: According to the `product_review` DB table, the `salesChannelId` is nullable, so we could just migrate it with a `null` value I think 🤔

src/Profile/Shopware6/Converter/SalesChannelDomainConverter.php
```diff
@@ -38,10 +38,20 @@ protected function convertData(array $data): ConvertStruct
         $converted['snippetSetId'] = $this->getMappingIdFacade(DefaultEntities::SNIPPET_SET, $data['snippetSetId']);
         $converted['salesChannelId'] = $this->getMappingIdFacade(DefaultEntities::SALES_CHANNEL, $data['salesChannelId']);
 
+        if ($converted['salesChannelId'] === null) {
+            return new ConvertStruct(null, $data);
```
MalteJanz, line 52: Because the `sales_channel_domain` entity is so closely related to sales channels, and otherwise useless, I would be fine with making an exception here and not logging it

src/Profile/Shopware6/Converter/SeoUrlConverter.php
```diff
@@ -80,6 +80,10 @@ protected function convertData(array $data): ConvertStruct
                 DefaultEntities::SALES_CHANNEL,
                 $converted['salesChannelId']
             );
+
+            if ($converted['salesChannelId'] === null) {
+                return new ConvertStruct(null, $data);
```
MalteJanz, line 94: `seo_url.sales_channe_id` is nullable, so we could technically still migrate it I guess. But we would need to take a closer look if that would make sense. Otherwise again we should log that this data wasn't migrated

src/Profile/Shopware6/Converter/SeoUrlTemplateConverter.php
```diff
@@ -38,8 +38,20 @@ protected function convertData(array $data): ConvertStruct
...
 
+        $mappedSalesChannelId = null;
+        if (isset($data['salesChannelId'])) {
+            $mappedSalesChannelId = $this->getMappingIdFacade(
+                DefaultEntities::SALES_CHANNEL,
+                $data['salesChannelId']
+            );
+
+            if ($mappedSalesChannelId === null) {
+                return new ConvertStruct(null, $data);
```
MalteJanz, line 59: see above, technically also nullable here

src/Profile/Shopware6/Converter/SystemConfigConverter.php
```diff
@@ -38,7 +38,19 @@ protected function convertData(array $data): ConvertStruct
...
-        $systemConfigUuid = $this->systemConfigLookup->get($data['configurationKey'], $data['salesChannelId'] ?? null, $this->context);
+        $mappedSalesChannelId = null;
+        if (isset($data['salesChannelId'])) {
+            $mappedSalesChannelId = $this->getMappingIdFacade(
+                DefaultEntities::SALES_CHANNEL,
+                $data['salesChannelId']
+            );
+
+            if ($mappedSalesChannelId === null) {
+                return new ConvertStruct(null, $data);
```
MalteJanz, line 59: see above

UPGRADE.md
```diff
@@ -1,4 +1,15 @@
 # NEXT
+- [BREAKING] [#178](https://github.com/shopware/SwagMigrationAssistant/pull/178) - fix!: unsupported sales channel migration
+  - [BREAKING] Added required constructor parameter:
+    - `SwagMigrationAssistant\Migration\Mapping\MappingServiceInterface`
+    - `SwagMigrationAssistant\Migration\Logging\LoggingServiceInterface`
+    - `SwagMigrationAssistant\Migration\Mapping\Lookup\SalesChannelTypeLookup`
+    - `SwagMigrationAssistant\Migration\Mapping\Lookup\SalesChannelLookup`
+      to service `SwagMigrationAssistant\Profile\Shopware6\Converter\SalesChannelConverter`
+- Added new lookup services `SwagMigrationAssistant\Migration\Mapping\Lookup\SalesChannelLookup` and `SwagMigrationAssistant\Migration\Mapping\Lookup\SalesChannelTypeLookup`
+- Changed Shopware 6 sales channel migration to read all source sales channel types. Sales channels with unsupported target types are skipped and logged with `SwagMigrationAssistant\Migration\Logging\Log\ConvertObjectTypeUnsupportedLog`
```
MalteJanz, line 11: Would also add a user facing changelog that clearly explains the fix and that extra sales channels, like social shopping, are now properly migrate, as long as the types exist before the migration (e.g. you have to install social shopping beforehand)

MalteJanz (commented):
nit: please provide a better PR description than

> Improve migration of sales-channel for Shopware 6

How did it improve?
What approach did you choose to solve the mentioned issue?
Why did you choose that approach, where there alternatives?

src/Profile/Shopware6/Converter/SalesChannelConverter.php
DennisGarding: The special case is not the name, but the fact that it has the same default ID. That is why this channel is handled different:  A new `ID` and renamed it to make clear that it is the migrated channel.

src/Profile/Shopware6/Converter/ProductReviewConverter.php
DennisGarding: Why you want to migrate a review that is lost because the sales channel does not exist?

src/Profile/Shopware6/Converter/SeoUrlConverter.php
DennisGarding: Same here. This is lost data without the SalesChannel

src/Profile/Shopware6/Converter/SeoUrlTemplateConverter.php
DennisGarding: Same here. This is lost data without the SalesChannel

src/Profile/Shopware6/Converter/SalesChannelConverter.php
MalteJanz: Are you talking about the other if branch above, where we generate the new id? Thats the one I understand and also the one we had before.

I don't understand the highlighted (new) `elseif` condition and branch, which only does renaming as far as I can see:
https://github.com/shopware/SwagMigrationAssistant/blob/8be7f590ea309249770b6c82ddbb9c47db310403/src/Profile/Shopware6/Converter/SalesChannelConverter.php#L66-L71

Can you explain why that is needed?

src/Profile/Shopware6/Converter/ProductReviewConverter.php
MalteJanz: 🤷 someone put in effort writing a review for a product and maybe you want to make use of that in a different sales channel 🤷 

But interestingly the definition says the sales channel id is required, so not sure how the feature would work with a `null` value in it 😬 
https://github.com/shopware/shopware/blob/167a494a32c541bd0bffbf305fbcf11b9457ec38/src/Core/Content/Product/Aggregate/ProductReview/ProductReviewDefinition.php#L68

TLDR: maybe just add a log as in the other cases and then skip it.

src/Profile/Shopware6/Converter/SeoUrlConverter.php
MalteJanz: I don't know what the use case is for seo urls without assigned sales channels (would need to research that), but that is allowed in this case according to the DB and entity definition:
https://github.com/shopware/shopware/blob/167a494a32c541bd0bffbf305fbcf11b9457ec38/src/Core/Content/Seo/SeoUrl/SeoUrlDefinition.php#L50

src/Profile/Shopware6/Converter/SalesChannelConverter.php
DennisGarding: This prevents a migrated SalesChannel from being merged with an existing SalesChannel in the shop

jozsefdamokos (approved):

src/Profile/Shopware6/Converter/SalesChannelConverter.php
MalteJanz: I might be wrong here but in that case what happens if you migrate multiple times and reset the checksums in between? Would it always append the suffix and you end up with `Storefront (Migrated) (Migrated)` and so on? Would it add new sales channels every time?

I really don't think its a good idea to do the lookup by translated name. The old implementation only handled the edge case for the default storefront (which has a known hard coded id, so we have to replace that id once and because we can cleanly handle that edge case also add the suffix).

But for all other cases I think its totally fine to just migrate the sales channels as is, even if they end up with the same translated name (there is no unique constraint). So then the merchant has to figure out which is which but in most cases their target shop should be empty anyways and they don't run into this at all (besides the default sales channel).

MalteJanz (changes requested):

UPGRADE.md
```diff
@@ -1,5 +1,17 @@
 # 17.0.0
 
+- [BREAKING] [#178](https://github.com/shopware/SwagMigrationAssistant/pull/178) - fix!: unsupported sales channel migration
```
MalteJanz: All these new entries need to be moved up under a new section `# NEXT`, we released `17.0.0` last week ✌️

MalteJanz (approved):

UPGRADE.md
DennisGarding: Strange... I thought I does it

PR #182, opened 2026-04-23
fix: customer language mapping

larskemper:
resolves #SWAG-327674 support ticket

### Problem

Fixes a Shopware 5 customer migration issue where `languageId` could be missing even though the customer locale was already known in the same migration run.

`CustomerConverter` resolved the customer language only through `LanguageLookup`, which checks existing SW6 language entities. In affected runs, a language mapping for the locale already existed, but the target `language` entity had not been persisted yet. That left `languageId` unset.

### Changes

- add a shared Shopware converter helper that resolves a language by: existing language mapping for locale & fallback to `LanguageLookup`
- switch `CustomerConverter` to use this helper for `customerlanguage.locale`

### DIFF
--- a/src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
+++ b/src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
@@ -81,7 +81,7 @@ class LanguageReader extends AbstractReader implements ReaderInterface
     }
 
     /**
-     * @return list<mixed>
+     * @return list<string>
      */
     private function fetchCustomerLocaleIds(MigrationContextInterface $migrationContext): array
     {
@@ -89,6 +89,7 @@ class LanguageReader extends AbstractReader implements ReaderInterface
         $query = $connection->createQueryBuilder();
         $query->from('s_user', 'customer');
         $query->addSelect('customer.language');
+        $query->distinct();
         $query->where('customer.language IS NOT NULL');
 
         return $query->executeQuery()->fetchFirstColumn();
--- a/src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
+++ b/src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
@@ -40,7 +40,10 @@ class LanguageReader extends AbstractReader implements ReaderInterface
 
     public function read(MigrationContextInterface $migrationContext): array
     {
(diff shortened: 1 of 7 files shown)


larskemper:
@MalteJanz I'm not fully sure whether the previous "LanguageLookup only, no mapping lookup" behavior was intentional?

DennisGarding (approved):

jozsefdamokos (approved):

src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
```diff
@@ -76,6 +80,21 @@ private function fetchShopLocaleIds(MigrationContextInterface $migrationContext)
         return $query->fetchFirstColumn();
     }
 
+    /**
+     * @return list<string>
+     */
+    private function fetchCustomerLocaleIds(MigrationContextInterface $migrationContext): array
```
MalteJanz, line 86: Is there a corresponding SW5 Migration Connector PR as well? The DB Readers for SW5 basically should use the same SQL in the assistant and connector plugin and we need to make sure they stay in sync (otherwise you get a Bug e.g. with the API gateway but not with the local gateway).

MalteJanz (approved):

src/Profile/Shopware/Gateway/Local/Reader/LanguageReader.php
larskemper: https://github.com/shopware/SwagMigrationConnector/pull/10

MalteJanz:
@larskemper just saw your question, but a late answer is still better than no answer I guess ✌️ 🙈 

> [Malte Janz (@MalteJanz)](https://github.com/MalteJanz) I'm not fully sure whether the previous "LanguageLookup only, no mapping lookup" behavior was intentional?

No I don't think that was intentional and we likely overlooked that we have to do both in most places 😬 
Now I'm also wondering if we broke other things with the introduction of these LookupServices as well or if languages are the exception. I mean we should only have this issue if it's an entity we migrate as well as lookup 🤔

larskemper:
> ... wondering if we broke other things with the introduction of these LookupServices as well or if languages are the exception. I mean we should only have this issue if it's an entity we migrate as well as lookup 🤔

Makes sense, thanks. Not sure either whether languages are the only affected case, so we should probably take a look at the other LookupServices as well.

PR #184, opened 2026-04-24
fix: consider time zone during migration of orders

### LINKED ISSUE #14186: Timezone differences between source and target systems
### Shopware Version

all

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

When migrating from a shop with a different timezone from UTC, the timestamps are migrated as they are, but in shopware 6 they are treated as if they were UTC timezone. This leads to orders having different timestamps in the source and target systems.

### Expected behaviour

The times should match between source and target systems

### How to reproduce

- have a source system (SW5) with different timezone setup from UTC
- migrate to SW6

DennisGarding:
closes: [#14184](https://github.com/shopware/shopware/issues/14186)

Gets time zone from source db and consider it during migration of orders

### DIFF
--- a/src/Profile/Shopware/Converter/OrderConverter.php
+++ b/src/Profile/Shopware/Converter/OrderConverter.php
@@ -55,6 +55,10 @@ abstract class OrderConverter extends ShopwareConverter
 
     private const SHIPPING_ADDRESS = 'shipping';
 
+    private const SOURCE_TIMEZONE_FIELD = '_timezone';
+
+    private const UTC_TIMEZONE = 'UTC';
+
     protected string $mainLocale;
 
     protected Context $context;
@@ -187,7 +191,7 @@ abstract class OrderConverter extends ShopwareConverter
         ];
         $converted['totalRounding'] = $converted['itemRounding'];
 
-        $this->convertValue($converted, 'orderDateTime', $data, 'ordertime', self::TYPE_DATETIME);
+        $this->convertOrderDateTime($converted, $data);
 
         if (isset($data['status'])) {
             $stateMapping = $this->mappingService->getMapping(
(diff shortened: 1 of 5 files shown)


src/Profile/Shopware/Converter/OrderConverter.php
```diff
@@ -1003,6 +1007,56 @@ protected function getSalutation(string $salutation): ?string
         return $salutationMapping['entityId'];
     }
 
+    /**
+     * @param array<string, mixed> $converted
+     * @param array<string, mixed> $data
+     */
+    private function convertOrderDateTime(array &$converted, array &$data): void
```
larskemper, line 1014: are there more datatime fields that would need something like that?

src/Profile/Shopware/Gateway/Local/Reader/OrderReader.php
```diff
@@ -151,12 +151,16 @@ private function appendAssociatedData(array $orders, MigrationContextInterface $
         $orderEsd = $this->getOrderEsd($migrationContext);
         $orderDetails = $this->getOrderDetails($migrationContext);
         $orderDocuments = $this->getOrderDocuments($migrationContext);
+        $timezone = $this->getDatabaseTimezone($migrationContext);
```
jozsefdamokos, line 154: Is it not still possible to have a mismatch here? Is the database timezone always aligned with the source shop's timezone?

src/Profile/Shopware/Converter/OrderConverter.php
```diff
@@ -1003,6 +1007,56 @@ protected function getSalutation(string $salutation): ?string
...
+            unset($data['ordertime'], $data[self::SOURCE_TIMEZONE_FIELD]);
+
+            return;
+        }
+
+        try {
+            $orderDateTime = new \DateTimeImmutable($data['ordertime'], new \DateTimeZone($sourceTimezone));
+            $converted['orderDateTime'] = $orderDateTime
+                ->setTimezone(new \DateTimeZone(self::UTC_TIMEZONE))
+                ->format(\DATE_ATOM);
```
MalteJanz, line 1035: Shouldn't this be `Defaults::STORAGE_DATE_TIME_FORMAT` instead, because thats what SW6 uses right?

src/Profile/Shopware/Converter/OrderConverter.php
MalteJanz: Yes I also think there are likely more cases we have to adjust. Just searching through the migration-assistant codebase for `datetime` in PHP files already reveals some more...

Looks like basically any time we call
```php
$this->convertValue(..., self::TYPE_DATETIME);
// or
$this->convertValue(..., 'datetime');
```
it should consider the source system timezone instead of assume its UTC as SW6. Means we need a more general proper fix for this instead of just fixing this one place for order times.

src/Profile/Shopware/Gateway/Local/Reader/AbstractReader.php
```diff
@@ -156,6 +156,32 @@ final protected function getDefaultShopLocale(MigrationContextInterface $migrati
...
+                <<<'SQL'
+                    SELECT timeZones.timeZone
+                    FROM (
+                        SELECT @@SESSION.time_zone AS timeZone
+                        UNION
+                        SELECT @@system_time_zone AS timeZone
+                    ) AS timeZones
+                    WHERE timeZone != 'SYSTEM'
+                    LIMIT 1
+                SQL
```
MalteJanz, line 172: Is this the most reliable way to figure out the storage timezone for SW5?
Because in my local dev system running this query on either my SW6 or SW5 database, both return `CEST`, while SW6 for example clearly stores all datetimes in UTC, means the DB timezone might not matter at all for the application?
Is there maybe some configuration value in SW5 to change it? Or does SW5 store its values with the timezone? I think in SW6 its programmatically hard coded in multiple places like:

https://github.com/shopware/shopware/blob/31828222cf8ef317198f9c3a0e568d3443479102/src/Core/Framework/DataAbstractionLayer/FieldSerializer/DateTimeFieldSerializer.php#L52-L54

src/Profile/Shopware/Gateway/Local/Reader/OrderReader.php
```diff
@@ -151,12 +151,16 @@ private function appendAssociatedData(array $orders, MigrationContextInterface $
...
 
         // represents the main language of the migrated shop
         $locale = $this->getDefaultShopLocale($migrationContext);
 
         foreach ($orders as &$order) {
             $order['_locale'] = \str_replace('_', '-', $locale);
+            if ($timezone !== null) {
+                $order['_timezone'] = $timezone;
```
MalteJanz, line 162: This looks wrong, why would you add this timezone data to each order while it is the same for every order, or even every entity that comes from SW5.

You should read that once from the source system and store it e.g. in a mapping for easy lookup 🤔

MalteJanz (changes requested):

src/Profile/Shopware/Gateway/Local/Reader/OrderReader.php
DennisGarding: We have 2 settings here:

PHP TimeZone: `phpsettings.date.timezone`
MySQL TimeZone: `db.timezone`

But Shopware 5 writes `s_order.ordertime` in the core using MySQL NOW(), not PHP DateTime.

src/Profile/Shopware/Gateway/Local/Reader/AbstractReader.php
DennisGarding: We could use `db.timezone` from the config and as fallback the PHP time zone

DennisGarding:
Closed: prefer a different implementation

PR #186, opened 2026-05-04
fix!: read timezone from config

### LINKED ISSUE #14186: Timezone differences between source and target systems

DennisGarding:
closes: https://github.com/shopware/shopware/issues/14186

This PR adds a source timezone premapping for Shopware migrations and uses it when converting date/time values.

The source timezone is read automatically only for Shopware API gateway connections. In that case, the Migration Connector endpoint returns the source system timezone, and we preselect it if it is a valid PHP timezone identifier.

For local gateway connections, the timezone is not read automatically. The premapping is still displayed so users can select or retain a timezone manually, but we do not call the API timezone reader for local connections. This avoids using the incorrect transport for local gateways and prevents crashes on connections that do not use the API connector.

If no valid source timezone can be detected, the migration retains the existing behavior unless the user explicitly configures a timezone in the premapping.


src/Profile/Shopware/Premapping/TimezoneReader.php
```diff
@@ -0,0 +1,124 @@
...
+class TimezoneReader extends AbstractPremappingReader
+{
+    private const MAPPING_NAME = 'source_timezone';
+
+    private const SOURCE_ID = 'timezone';
+
+    private array $validTimezones;
+
+    public function __construct(
+        private readonly ApiTimezoneReader $apiTimezoneReader,
```
DennisGarding: I am currently unsure how to implement it. Inject both services or add them to the registry.

If i should add them to the registry i need a bulletproof `supports function`  for the reader and currently possible is only 

`$migrationContext->getDataSet() === null`

Maybe anyone has an idea?

src/Profile/Shopware/Gateway/Local/Reader/TimezoneReader.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        $basePath = $fields['installationRoot'];
+
+        $configFile = \rtrim($basePath, '/\\') . '/config.php';
+        if (!\is_file($configFile) || !\is_readable($configFile)) {
+            return [];
+        }
+
+        $swConfig = include $configFile;
```
MalteJanz: from a security standpoint, this might be a bit dangerous 😕 

An alternative idea would be just trying to read the text and run a Regex over it to find the timezone value

src/Profile/Shopware/Premapping/TimezoneReader.php
MalteJanz: I would prefer using the `ReaderRegistry` and looking up the concrete reader that way. But you're right, you need a `supports` method that returns true only for this case, otherwise it would cause issues down the line and might be returned in the wrong places.

What speaks against introducing a "virtual" entity, by introducing a `DataSet` that is isn't part of any `DataSelection` group, so the user can't select it for the migration but we can still use it for premapping lookup? Might be worth a try so you can compare against it in your `supports` method 🤔 

An alternative could also be looking into the `EnvironmentReader`, where we also read out stuff like the `defaultCurrency` of the source system and if it could fit somewhere in there 🤷

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +192,48 @@ protected function getAttributes(
...
+        }
+
+        $connectionId = $this->migrationContext->getConnection()->getId();
+        if (!isset($this->timezoneCache[$connectionId])) {
+            $this->timezoneCache[$connectionId] = $this->mappingService->getValue(
+                $connectionId,
+                'source_timezone',
+                'timezone',
+                $context
+            ) ?? '';
```
jozsefdamokos: I dont think it makes much sense setting it to an empty string if the mapping does not exist. We can also return the datetime with default timezone right away

src/Profile/Shopware/Gateway/Local/Reader/EnvironmentReader.php
```diff
@@ -12,24 +12,53 @@
 use Shopware\Core\Framework\Log\Package;
 use SwagMigrationAssistant\Migration\Gateway\Reader\EnvironmentReaderInterface;
 use SwagMigrationAssistant\Migration\MigrationContextInterface;
+use Symfony\Component\Filesystem\Filesystem;
 
+/**
+ * @phpstan-type ReadArray array{defaultShopLanguage: string, host: string, additionalData: array<int, mixed>, defaultCurrency: string, config: array<string, mixed>, timezone: string}
```
jozsefdamokos: Maybe something like`EnvironmentInfo` instead of `ReadArray` fits better

src/Profile/Shopware/Gateway/Local/Reader/EnvironmentReader.php
```diff
@@ -120,4 +149,75 @@ private function getAdditionalData(MigrationContextInterface $migrationContext):
 
         return \array_values($shops);
     }
+
+    private function getTimezone(MigrationContextInterface $migrationContext): string
```
jozsefdamokos: Maybe makes sense to return `string` or `null` in case timezone cant be found so `: ?string`

jozsefdamokos (approved):

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -23,10 +24,16 @@ abstract class ShopwareConverter extends Converter
...
     protected const TYPE_FLOAT = 'float';
+    protected const TYPE_DATE = 'date';
     protected const TYPE_DATETIME = 'datetime';
 
     protected MigrationContextInterface $migrationContext;
 
+    /**
+     * @var array<string, string|null>
+     */
+    protected array $timezoneCache = [];
```
larskemper: Services with cache should implement `ResetInterface`. Also not sure if thats the right "place" here, maybe a dedicated service

src/Profile/Shopware/Gateway/Api/Reader/EnvironmentReader.php
```diff
@@ -28,6 +28,11 @@ class EnvironmentReader implements EnvironmentReaderInterface
 {
     private ?HttpClientInterface $client = null;
 
+    /**
+     * @var array<string, ReadArray>
+     */
+    private array $cachedEnvironmentInformation = [];
```
larskemper: Same here

tests/acceptance/fixtures/shopware5-source/config.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types=1);
```
larskemper: is this needed? I would prefer to not mixin php to the acceptance code base

tests/acceptance/tests/MigrationTest.spec.ts
```diff
@@ -75,6 +75,11 @@ test.describe('Migration Tests @migration @visual', () => {
                 }
             }
 
+            await page
```
larskemper: why is that needed? the code above should walk on it one to the end of the tab listing

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -79,12 +87,23 @@ protected function convertValue(
...
-                case self::TYPE_DATETIME:
+                case self::TYPE_DATE:
                     $sourceValue = $sourceData[$sourceKey];
                     if (!$this->validDate($sourceValue)) {
                         return;
                     }
 
+                    break;
+                case self::TYPE_DATETIME:
+                    if (!$context instanceof Context) {
```
larskemper: is this a break? previously datetime worked without a context, now its just dropped when there is non. Maybe logging this case makes sense?

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +192,48 @@ protected function getAttributes(
 
         return $result;
     }
+
+    private function validDate(string $value): bool
```
larskemper: was `protected` before

src/Profile/Shopware/Gateway/Local/Reader/EnvironmentReader.php
```diff
@@ -120,4 +149,75 @@ private function getAdditionalData(MigrationContextInterface $migrationContext):
...
+    private function readDbTimezoneFromConfig(string $phpFileContent): ?string
+    {
+        $tokens = token_get_all($phpFileContent);
+        $inDbArray = false;
+        $arrayDepth = 0;
+
+        foreach ($tokens as $index => $token) {
+            $value = \is_array($token) ? \trim($token[1], self::TRIM_QUOTATION) : $token;
+
+            if (!$inDbArray && $value === self::DB_KEY) {
```
larskemper: Isnt this matching any value = "db". So for example also `"note" => "db"` where its not a key.

src/DependencyInjection/shopware.php
```diff
@@ -169,6 +169,7 @@
 use SwagMigrationAssistant\Profile\Shopware\Premapping\OrderStateReader;
 use SwagMigrationAssistant\Profile\Shopware\Premapping\PaymentMethodReader;
 use SwagMigrationAssistant\Profile\Shopware\Premapping\SalutationReader;
+use SwagMigrationAssistant\Profile\Shopware\Premapping\TimezoneReader as PremappingTimezoneReader;
```
MalteJanz: Nit: whats the reason behind this alias? I think `TimezoneReader` is already unique in this file? And all other premapping readers also don't have a "Premapping" prefix 🤔

src/Profile/Shopware/Converter/ShopwareConverter.php
MalteJanz: Yes I would also consider that a break, which should be documented in the `upgrade.md`.

And yes we should not silently return and do nothing here. I would say calling this `convertValue` method with a type of `TYPE_DATETIME` and without a `Context` should be considered a developer error, means you used the API wrong, which should produce some obvious error for the developer.

In our case I guess its best to use our logging system to create a non-user-fixeable error with details like a stack trace to find out where this was wrongly called from (could also be from another plugin like SwagMigrationMagento)

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +192,48 @@ protected function getAttributes(
...
+
+    private function convertDateTime(string $value, Context $context): ?string
+    {
+        if ($value === '') {
+            return null;
+        }
+
+        $connectionId = $this->migrationContext->getConnection()->getId();
+        if (!\array_key_exists($connectionId, $this->timezoneCache)) {
+            $this->timezoneCache[$connectionId] = $this->mappingService->getValue(
```
MalteJanz: In addition to what Lars wrote that its not a great idea to have a cache without reset interface in this parent class:

We are only ever dealing with a single connection during the whole migration process. Means there is no point in caching values "per connection", so your cache could just be a `?string` instance variable instead of an associative array

MalteJanz: Additionally, if you consider this lookup only needs to happen once per PHP process / request, I'm not even sure if its worth it to put the `Context` in the method signatures just for that edge case.

E.g. you could also run this lookup once elsewhere, like a dedicated service and in it's constructor, and every `convertDateTime` then just uses that value and consequently you don't need to pass in the `Context` at all, means you also don't need error handling when the developer forgots to pass it and its not really a breaking change anymore 🤔

src/Profile/Shopware/Gateway/Api/Reader/EnvironmentReader.php
```diff
@@ -37,6 +42,10 @@ public function __construct(private readonly ConnectionFactoryInterface $connect
      */
     public function read(MigrationContextInterface $migrationContext): array
     {
+        if (isset($this->cachedEnvironmentInformation[$migrationContext->getConnection()->getId()])) {
+            return $this->cachedEnvironmentInformation[$migrationContext->getConnection()->getId()];
```
MalteJanz: I would vote against caching the environment information like this. Often you want to explicitly reach out to the source system again, e.g. to check if its now available with the given credentials and this caching might cause unnoticed bugs.

Additionally you really don't want to call this other than in the initial connection check, because its rather expensive to reach out the the source system and gather all this environment data.

src/Migration/EnvironmentInformation.php
```diff
@@ -32,6 +32,7 @@ public function __construct(
         protected string $sourceSystemLocale = '',
         protected string $targetSystemLocale = '',
         protected ?string $fingerprint = null,
+        protected ?string $timezone = null,
```
MalteJanz: Nit: considering the other variable names I would suggest a more specific name here as well:

```suggestion
        protected ?string $sourceSystemStorageTimezone = null,
```
or just
```suggestion
        protected ?string $sourceSystemTimezone = null,
```

src/Profile/Shopware/Gateway/Local/Reader/EnvironmentReader.php
MalteJanz: In addition to that, there is also no exit condition when the loop runs outside the `'db' => []` array again, so it could still match a `timzone` under a different parent key than `db`.

To be honest I find this code quite complex and I'm not sure if we need this complexity for the convenience to set a default value instead of letting the user input that. Are there multiple `'timezone'` keys in a usual shopware 5 config file?

If not I would suggest just sticking to a simple Regex and if that doesn't find anything leave it up to the user. [Something like this](https://regex101.com/?regex=%5B%27%22%5Dtimezone%5B%27%22%5D%5Cs*%3D%3E%5Cs*%5B%27%22%5D%28.%2B%29%5B%27%22%5D&testString=%3C%3Fphp+declare%28strict_types%3D1%29%3B%0A%2F*%0A+*+%28c%29+shopware+AG+%3Cinfo%40shopware.com%3E%0A+*+For+the+full+copyright+and+license+information%2C+please+view+the+LICENSE%0A+*+file+that+was+distributed+with+this+source+code.%0A+*%2F%0A%0Areturn+%5B%0A++++%27db%27+%3D%3E+%5B%0A++++++++%27timezone%27+%3D%3E+%27Europe%2FBerlin%27%2C%0A++++%5D%2C%0A%5D%3B%0A%0A%2F%2F+test+some+different+config%0Areturn+%5B%0A++++%27db%27+%3D%3E+%5B%0A++++++++%27timezone%27+%3D%3E+%0A++++++++++++++++++++%27Europe%2FBerlin%27%2C%0A++++%5D%2C%0A%5D%3B%0A%0A%2F%2F+another+variation%0Areturn+%5B%0A++++%27db%27+%3D%3E+%5B%0A++++++++%22timezone%22+%3D%3E+%0A++++++++++++++++++++%22Europe%2FBerlin%22%2C%0A++++%5D%2C%0A%5D%3B%0A&flags=gm&flavor=pcre2&delimiter=%2F)


```regex
['"]timezone['"]\s*=>\s*['"](.+)['"]
```

And if that doesn't return a valid timezone you wouldn't show that as a suggestion in the premapping anyways right?

MalteJanz (commented):
The more I think about it, the more I think the extra effort for reading the SW5 configured timezone isn't worth it. Sorry for leading you down this `EnvironmentReader` path with [my previous comment](https://github.com/shopware/SwagMigrationAssistant/pull/186#discussion_r3187022031) and our short talk.

What if we apply the simplest solution we could come up with here, just a few of my thoughts (feel free to explore a simpler approach on your own / deviate from this):
- No reading (or guessing) from the source system at all
- Premapping has a option for setting the "storage timezone" of the source system, which is like all other premappings put into the mapping table and can be used from there
- possible choices would be all valid timezones
- we could think about defaulting / pre-selecting "UTC", if that works for the majority of cases and otherwise the user can still adjust it and migrate again
- I still like your approach of using `Converter->convertValue(..., TYPE_DATETIME)` to do the correct conversion automatically, but I would prefer if we could get rid of the extra `Context` parameter there and initialize the internal used source timezone differently from the mapping

My reasoning behind this:
- Doing a full "read environment" call is expensive and should not really be necessary. I previously thought we might store the values of that somewhere in the DB but I was wrong about that (turns out we store it on the `swag_migration_run` but at the premapping stage we don't have a run yet)
- The other premapping readers also sometimes read from the source system, but they only use a dedicated `$gateway->readTable` API. In our case we don't need DB data but rather configuration values stored as a file. Introducing extra APIs for that seems not worth the effort
- Reading the SW5 config from a PHP file is error prone, e.g. they could also store the timezone in a variable and use that in multiple places `'timezone' => $timezone` or other strange things
- (Not sure about this one) but is that `config.php` really the single source of truth for this configuration in SW5? Is it not possible to override / provide config values via environment variables or other ways, where this `config.php` value could be wrong / unused?
- The only reliable way I see would be asking the SW5 backend itself (e.g. they executed `config.php` and can tell what value is "active"), but that wouldn't work for our local gateway and doesn't seem worth the effort just for giving the user a little more convenience

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +195,71 @@ protected function getAttributes(
...
+            }
+
+            $date = new \DateTimeImmutable($value, new \DateTimeZone($timezone));
+
+            return $date
+                ->setTimezone(new \DateTimeZone('UTC'))
+                ->format(Defaults::STORAGE_DATE_TIME_FORMAT);
+        } catch (\Throwable $exception) {
+            $this->loggingService->log(
+                MigrationLogBuilder::fromMigrationContext($this->migrationContext)
```
larskemper: would be good to add at least entityName here, so its possible to identify

src/Migration/Logging/Log/ConvertDateTimeFailedLog.php
```diff
@@ -0,0 +1,30 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace SwagMigrationAssistant\Migration\Logging\Log;
+
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Migration\Logging\Log\Builder\AbstractMigrationLogEntry;
+
+#[Package('fundamentals@after-sales')]
+readonly class ConvertDateTimeFailedLog extends AbstractMigrationLogEntry
```
larskemper, line 14: Would not add a new Log for this. In the end it can simply be a `ConvertEntityFailedLog` or `RunExceptionLog`

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +195,71 @@ protected function getAttributes(
...
+    {
+        $runId = $this->migrationContext->getRunUuid();
+        if (\array_key_exists($runId, $this->timezoneCache)) {
+            return $this->timezoneCache[$runId];
+        }
+
+        $timezone = null;
+        $premapping = $this->migrationContext->getConnection()->getPremapping();
+        foreach ($premapping ?? [] as $item) {
+            if ($item->getEntity() !== 'source_timezone') {
```
larskemper: would be good to make constants in `TimezoneReader` public and reuse here

larskemper (approved):

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -174,4 +195,71 @@ protected function getAttributes(
...
+                continue;
+            }
+
+            foreach ($item->getMapping() as $mapping) {
+                if ($mapping->getSourceId() === 'timezone') {
+                    $timezone = $mapping->getDestinationUuid() === '' ? null : $mapping->getDestinationUuid();
+                }
+            }
```
MalteJanz, line 283: This is not how we read the submitted premapping in our converters, see our documentation on that here:
https://developer.shopware.com/docs/products/extensions/migration-assistant/concept/premapping.html

You just need to do a mapping lookup instead, e.g.:
```php
$mapping = $this->mappingService->getMapping(
            $this->connectionId,
            TimezoneReader::getMappingName(),
            TimezoneReader::SOURCE_ID,
            $this->context,
        );
```

*You likely need to use `getValue` in this case though, because you don't want to look up a UUID, but instead just a string
https://github.com/shopware/SwagMigrationAssistant/blob/d970390e49d13036c09af246a11d74a6e5e7927a/src/Migration/Mapping/MappingService.php#L260

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -7,26 +7,41 @@
...
 use SwagMigrationAssistant\Migration\DataSelection\DefaultEntities;
+use SwagMigrationAssistant\Migration\Logging\Log\Builder\MigrationLogBuilder;
+use SwagMigrationAssistant\Migration\Logging\Log\ConvertDateTimeFailedLog;
 use SwagMigrationAssistant\Migration\Mapping\Lookup\LanguageLookup;
 use SwagMigrationAssistant\Migration\MigrationContextInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 #[Package('fundamentals@after-sales')]
-abstract class ShopwareConverter extends Converter
+abstract class ShopwareConverter extends Converter implements ResetInterface
```
MalteJanz, line 25: Where are all the child classes tagged with `kernel.reset` now? Otherwise just implementing this Interface is useless if its never called.

You might want to extract that datetime timezone conversion logic into its own service though, that way its stays independent of the `ShopwareConverter` and could potentially also be used by other profile converters such as Magento 🤷

src/Profile/Shopware/Premapping/TimezoneReader.php
```diff
@@ -0,0 +1,115 @@
...
+        );
+    }
+
+    private function readSourceTimezone(MigrationContextInterface $migrationContext): ?string
+    {
+        if ($migrationContext->getConnection()->getGatewayName() !== ShopwareApiGateway::GATEWAY_NAME) {
+            return null;
+        }
+
+        $timezoneResult = $this->timezoneReader->read($migrationContext);
```
MalteJanz, line 95: You can't just use the `ApiTimezoneReader` here, you need to go through the `GatewayRegistry` to get the correct gateway for the active connection and the use that to talk to the source system, otherwise this might just crash for connections with local gateway. Just take a look at our existing Premapping readers or our documentation about creating one:
https://developer.shopware.com/docs/products/extensions/migration-assistant/guides/decorating-a-shopware-migration-assistant-converter.html#creating-a-premapping-reader

But as I already said in my last code review, I don't think its worth the effort to read from the source system if that isn't even guaranteed to be reliable. Then its better to just let the user choose and maybe default to UTC if that's commonly set on SW5 by default?

src/Profile/Shopware/Premapping/TimezoneReader.php
```diff
@@ -0,0 +1,115 @@
...
+
+    public function getPremapping(Context $context, MigrationContextInterface $migrationContext): PremappingStruct
+    {
+        $this->fillConnectionPremappingDictionary($migrationContext);
+
+        $choices = $this->getTimeZoneList();
+
+        $sourceTimezone = $this->readSourceTimezone($migrationContext);
+
+        $description = $sourceTimezone ?? 'No source time zone';
```
MalteJanz: How does this description make sense? It should always say `Source system time zone`, not `Europe/Berlin` and then you have to choose `Europe/Berlin` as well 😕  ?

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
```diff
@@ -0,0 +1,25 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace SwagMigrationAssistant\Profile\Shopware\Gateway\Api\Reader;
+
+use Shopware\Core\Framework\Log\Package;
+use SwagMigrationAssistant\Migration\MigrationContextInterface;
+
+#[Package('after-sales')]
+class TimezoneReader extends ApiReader
```
MalteJanz, line 26: If you really want to do this reading from the source system, where is the equivalent local gateway reader? Also the concept of these Readers are for reading data that is then converted, and are not really for fetching premapping data. For that all other PremappingReaders are using the `GatewayInterface->readTable` method.

I would again strongly recommend not to fetch the timezone from the source system, unless you find a clean solution that is guaranteed to work and always provides the correct source system timezone for both API and local gateways

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function supports(MigrationContextInterface $migrationContext): bool
+    {
+        return false;
```
MalteJanz: Dead code? Directly using this Reader from dependency injection is an anti-pattern and it should only be retrieved from the corresponding Registry, to support all possible gateways

MalteJanz (changes requested):
Because there aren't any comments / reactions to my previous comments (other than that they are resolved) and there is no meaningful PR description I have no clue what your current goal for a solution is. Looks like you still want to read the source system timezone?

Nevertheless the current implementation still has lots of issues related to that and also didn't follow my suggestion to not read from the source system to keep it simple. Please provide proper solutions to these issues or refactor to not read from the source system.

src/Profile/Shopware/Converter/ShopwareConverter.php
DennisGarding: You’re absolutely right, of course, but unfortunately there isn’t an entity available here.

src/Migration/Logging/Log/ConvertDateTimeFailedLog.php
DennisGarding: As I don’t have an entity available, it doesn’t really fit, at least not in terms of the name. I’d actually leave it as it is to make it easier to identify later on, as we have our own code.

src/Profile/Shopware/Converter/ShopwareConverter.php
DennisGarding: In that case, I do need the context again... In your last review, you said I should remove them :man_shrugging:

src/Profile/Shopware/Converter/ShopwareConverter.php
larskemper: it is: https://github.com/shopware/SwagMigrationAssistant/blob/a46760d9de1304daa9d6db5eaebbc5b189b79ef4/src/Profile/Shopware/Converter/ShopwareConverter.php#L127

src/Profile/Shopware/Converter/ShopwareConverter.php
MalteJanz: I meant removing the context from the interface / callsite, e.g. you could:
- extract this logic into its own service
- run some "setup" / "init" code once, which takes the context as an argument and caches the result of the lookup
- afterwards you can always read the cached value without providing a context

Maybe it also fits well into an existing service somewhere. In the end we still have to do the mapping lookup (even better would be removing the context from that, just using plain SQL queries for performance, but that would be huge refactor).

src/Profile/Shopware/Converter/ShopwareConverter.php
DennisGarding: > Where are all the child classes tagged with kernel.reset now? Otherwise just implementing this Interface is useless if its never called.

Ah, you're right. I forgot to set the tag in the DI class. :+1: 

> You might want to extract that datetime timezone conversion logic into its own service though, that way its stays independent of the ShopwareConverter and could potentially also be used by other profile converters such as Magento 🤷

I think this is a very good place for this, as it specifically concerns the migration from SW5 to SW6. 

Magento generally stores the relevant technical timestamps in UTC, like SW6

Data in Magento 1–2 should not be converted to UTC again using the store time zone. In Magento, this would essentially amount to a double conversion and would corrupt the timestamps.

What I would nevertheless consider separately are semantically pure date fields (TYPE_DATE). A pull request for this will follow later.

src/Profile/Shopware/Premapping/TimezoneReader.php
DennisGarding: > You can't just use the ApiTimezoneReader here, you need to go through the GatewayRegistry to get the correct gateway for the active connection and the use that to talk to the source system, otherwise this might just crash for connections with local gateway. Just take a look at our existing Premapping readers or our documentation about creating one:

The current implementation should not crash for local gateway connections, because the ApiTimezoneReader is only called when the connection gateway name is `api`.

I see the architectural point. Existing premapping readers use the GatewayRegistry and the documented pattern. The reason for the direct reader here is that the timezone endpoint is API-specific and not represented by the local profile. 

> But as I already said in my last code review, I don't think its worth the effort to read from the source system if that isn't even guaranteed to be reliable. Then its better to just let the user choose and maybe default to UTC if that's commonly set on SW5 by default?

This is the most user-friendly solution. The time zone can be reliably retrieved in the API environment. In the local profile, `null` is returned, as we cannot reliably determine the time zone.

src/Profile/Shopware/Premapping/TimezoneReader.php
DennisGarding: Maybe it makes more sense for you if you can see the UI: 

image

The description is the value for `Previous value`. The text is discussed with @jc-sw-git during the Demo. 

If there is a `Previous value` i should use it.

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
DennisGarding: As I said, the time zone can be reliably retrieved in the API environment. In the local profile, null is returned, as we cannot reliably determine the time zone. But this is the most user-friendly solution.

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
DennisGarding: In principle, I would agree with you, but in this case, the API call is the simplest, most meaningful, and fastest solution. We should discuss whether the service should be placed elsewhere and whether it would make more sense to throw an exception in this method.

DennisGarding:
> Because there aren't any comments / reactions to my previous comments (other than that they are resolved) and there is no meaningful PR description I have no clue what your current goal for a solution is. Looks like you still want to read the source system timezone?

Oh sorry, i forgot the PR descriptioin. :sweat_smile: 

We had agreed that once we’d dealt with a comment, it would be marked as resolved. 
 
> Nevertheless the current implementation still has lots of issues related to that and also didn't follow my suggestion to not read from the source system to keep it simple. Please provide proper solutions to these issues or refactor to not read from the source system.

I don’t see any (lots of issues related), and a suggestion is just that – a suggestion, not a specific instruction. If that were the case, the ticket would have needed to be drafted more thoroughly. 


This solution is the best we can offer our users.

src/Profile/Shopware/Premapping/TimezoneReader.php
MalteJanz: The issue I see here is that this is not a traditional premapping thing, where:
- "previous value" is the value in the source system
- "new assignment" is the value in the target SW6 system

The new assignment in SW6 would technically always be UTC, because we convert to that timezone. So what the user chooses here is what the actual timezone in SW5 is instead of what it should be mapped to in SW6.

That's why I think its better to explicitly say under "description" / "previous value" that this is the source system timezone you have to choose. Otherwise if you might see "Europe/Berlin" as previous value, you might think you should choose "UTC" as new assignment, because thats what you want to have in SW6...

src/Profile/Shopware/Converter/ShopwareConverter.php
MalteJanz: As discussed in our call, lets keep this implementation but add a comment that clearly explains why we have chosen this approach over using the mappingService

src/Profile/Shopware/Converter/ShopwareConverter.php
DennisGarding: Sure, but this is the wrong method: 

https://github.com/shopware/SwagMigrationAssistant/blob/a46760d9de1304daa9d6db5eaebbc5b189b79ef4/src/Profile/Shopware/Converter/ShopwareConverter.php#L72
```
protected function convertValue(
        array &$newData,
        string $newKey,
        array &$sourceData,
        string $sourceKey,
        string $castType = self::TYPE_STRING,
    ): void {
```

Ah, i see: in https://github.com/shopware/SwagMigrationAssistant/blob/a46760d9de1304daa9d6db5eaebbc5b189b79ef4/src/Profile/Shopware/Converter/ShopwareConverter.php#L125 its also called. 

Yes, but we dont have it reliably. Should I provide it if it's there?  Like:

```
 private function convertDateTime(string $value, ?string $entity): ?string)
 {
 ...
 }
```

src/Profile/Shopware/Premapping/TimezoneReader.php
MalteJanz: As discussed in our call, using this is here is fine to provide a better UX by giving a pre-selection to the user and gated behind an early return, which just returns null for other (local) gateways.

We just have to make sure its clear that the `ApiTimezoneReader` doesn't follow our normal "Reader" concept and instead is a special case just used here for the premapping (more on that in another comment on that class)

src/Profile/Shopware/Premapping/TimezoneReader.php
MalteJanz: As discussed in our call, would be better to just always say something like "source time zone" as description, regardless of the detected time zone (which is also preselected on the right "new assignment")

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
MalteJanz: As discussed in our call, we should make it clear that this `Api/Reader/TimezoneReader` is not one of our usual "Readers" that run during the migration process and returned by the `ReaderRegistry` by:
- throwing an Exception in this `supports` method as it shouldn't be called / registered in the `ReaderRegistry`
- adding a comment above the class that explains the purpose of this service, that it explicitly only exists to fetch the source system timezone over API for the premapping, and why it extends `ApiReader`, to make use of the existing API client code and not re-implement it
- optionally moving it to a different namespace

So that future us can easily see why we have implemented it this way and that this is clearly different from our usual and documented reader concept

src/Profile/Shopware/Converter/ShopwareConverter.php
MalteJanz: As discussed in our call, lets keep this implementation in the `ShopwareConverter`

src/Profile/Shopware/Converter/ShopwareConverter.php
MalteJanz: One other idea: you have access to the migration Context right? 🤔 
https://github.com/shopware/SwagMigrationAssistant/blob/a46760d9de1304daa9d6db5eaebbc5b189b79ef4/src/Profile/Shopware/Converter/ShopwareConverter.php#L33

with that you could also retrieve the current `DataSet` and it's corresponding entity. 

*Sure that might not be totally accurate e.g. if you call `convertValue` on some nested child entity like `manufacturer`, but at least you know that you have to look into the `product` Converter for example. But on the other hand we already have the exception trace for that 🤷

DennisGarding: ok, good idea. I will change that

DennisGarding: I have now implemented both solutions. I use the dataset from the migrationContext. Because it can be null I use a fallback.

MalteJanz (approved):

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -7,26 +7,45 @@
...
 {
     protected const TYPE_STRING = 'string';
     protected const TYPE_BOOLEAN = 'bool';
     protected const TYPE_INVERT_BOOLEAN = 'invert_bool';
     protected const TYPE_INTEGER = 'int';
     protected const TYPE_FLOAT = 'float';
+    protected const TYPE_DATE = 'date';
     protected const TYPE_DATETIME = 'datetime';
 
+    private const CONVERT_VALUE_LOG_ENTITY_NAME = 'shopware_converter_convert_value_log_entity';
```
larskemper: the entityName of a log should always be either a valid entity definition name or null. So we should only set it if its fillable.

```php
if ($entityName !== null) {
    $log->withEntityName($entityName);
}
```

src/Exception/MigrationException.php
```diff
@@ -567,4 +569,14 @@ public static function mainVariantRelationMissingIdAndOrderNumber(): self
             'MainVariantRelation requires ID and order number.',
         );
     }
+
+    public static function readerRegistryUsageNotAllowed(string $class): self
```
larskemper, line 577: you can let phpstan check the class to be stricter

```suggestion
     /**
     * @param class-string<AbstractReader> $class
     */
    public static function readerRegistryUsageNotAllowed(string $class): self
```

larskemper (approved):

src/Profile/Shopware/Gateway/Api/Reader/TimezoneReader.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * This is intentionally not a regular migration reader and must not be resolved
+ * through the ReaderRegistry during the migration process. It
+ * only exists to fetch the source system timezone from the Migration Connector
+ * API in the premapping.
+ *
+ * The class extends ApiReader only to reuse the existing API client/request
+ * handling for connector endpoints instead of duplicating that logic here.
+ */
+#[Package('after-sales')]
```
vintagesucks:
```suggestion
#[Package('fundamentals@after-sales')]
```

src/Profile/Shopware/Converter/ShopwareConverter.php
```diff
@@ -7,26 +7,43 @@
...
     protected const TYPE_FLOAT = 'float';
+    protected const TYPE_DATE = 'date';
     protected const TYPE_DATETIME = 'datetime';
 
     protected MigrationContextInterface $migrationContext;
 
+    /**
+     * @var array<string, string|null>
+     */
+    protected array $timezoneCache = [];
```
vintagesucks:
```suggestion
    private array $timezoneCache = [];
```

vintagesucks (approved):

PR #190, opened 2026-05-20
fix!: update product.sales during migration

### LINKED ISSUE #12748: Migrated orders are not included in the ‘Sales’ field in the product overview.
### Shopware Version

6.7.2.2

### Affected area / extension

Platform(Default)

### Actual behaviour

The migrated orders are not included in the ‘Sales’ column in the product overview.  

This means that only sales per product after migration are taken into account; all orders that were created previously and migrated later are not counted.

### Expected behaviour

All sales of products are taken into account if they are available in the shop. Regardless of whether they have been migrated or not.

### How to reproduce

Create an order in Shopware 5 and then migrate the product and order.  
Then look at the Sales column in the product overview in the admin area and check if the sale from Shopware 5 has been taken into account → Sales are still incorrectly showing as 0.

DennisGarding:
Closes: shopware/shopware#12748

- Recalculates `product.sales` after order migration based on persisted live order line items.
- Covers remigration cases by updating both previously and currently affected products.

### DIFF
--- a/src/Migration/Service/ProductSalesUpdater.php
+++ b/src/Migration/Service/ProductSalesUpdater.php
@@ -1,0 +1,138 @@
+<?php declare(strict_types=1);
+/*
+ * (c) shopware AG <info@shopware.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace SwagMigrationAssistant\Migration\Service;
+
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
+use Shopware\Core\Checkout\Cart\LineItem\LineItem;
+use Shopware\Core\Checkout\Order\OrderStates;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\DataAbstractionLayer\Doctrine\RetryableQuery;
+use Shopware\Core\Framework\Log\Package;
+use Shopware\Core\Framework\Uuid\Uuid;
+
+#[Package('fundamentals@after-sales')]
+final readonly class ProductSalesUpdater
+{
+    public function __construct(
+        private Connection $connection,
+    ) {
+    }
+
+    /**
+     * @param array<string> $orderIds
+     *
+     * @return array<string>
+     */
+    public function getProductIdsForOrders(array $orderIds): array
+    {
+        $orderIds = $this->filterIds($orderIds);
+
+        if ($orderIds === []) {
+            return [];
+        }
+
+        $sql = <<<'SQL'
+            SELECT DISTINCT order_line_item.product_id
+            FROM order_line_item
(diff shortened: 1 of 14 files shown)


src/Migration/Service/ProductSalesUpdater.php
vintagesucks, line 1: Not related to this file, but the new behavior should be mentioned in the changelog, right?

vintagesucks (approved):

src/Migration/Service/ProductSalesUpdater.php
DennisGarding: Ah yes, your right!

src/Migration/Service/ProductSalesUpdater.php
```diff
@@ -0,0 +1,138 @@
...
+                    ON `order`.id = order_line_item.order_id
+                    AND `order`.version_id = order_line_item.order_version_id
+                    AND `order`.version_id = :liveVersion
+                INNER JOIN state_machine_state
+                    ON state_machine_state.id = `order`.state_id
+                WHERE order_line_item.product_id IN (:productIds)
+                    AND order_line_item.version_id = :liveVersion
+                    AND order_line_item.product_version_id = :liveVersion
+                    AND order_line_item.type = :lineItemType
+                    AND state_machine_state.technical_name != :cancelledState
```
larskemper: Are there other state we need to exclude? refunded, returned ...

larskemper (approved):

src/Migration/Writer/OrderWriter.php
```diff
@@ -21,6 +24,7 @@ public function __construct(
         EntityWriterInterface $entityWriter,
         EntityDefinition $definition,
         private readonly StructNormalizer $structNormalizer,
+        private readonly ProductSalesUpdater $productSalesUpdater,
```
larskemper, line 27: Break. So please also update the `UPGRADE.md` and use bang notation for the commit: `fix!: ...`

src/Migration/Writer/OrderWriter.php
```diff
@@ -44,6 +48,84 @@ public function writeData(array $data, Context $context): array
         }
         unset($item);
 
-        return parent::writeData($data, $context);
+        // Re-migration case: product A is only visible before a line item switches to product B.
```
jozsefdamokos: nit: could be a private method with a more detailed comment. for me its not obvious what the code does

jozsefdamokos (approved):

src/Migration/Service/ProductSalesUpdater.php
DennisGarding: Yes, your right. now only open and completed counts

src/Migration/Writer/OrderWriter.php
DennisGarding: Added some more info

PR #192, opened 2026-05-25
fix: defer late foreign key links

### LINKED ISSUE #10609: SW6->SW6: Order line item `promotionId` prevents migration of Orders if set
The {{promotionId}} field is a foreign key on the {{promotion}} table and thus the referenced Id’s need to be there (which currently isn’t the case when “promotions” aren’t checked for migration). If you check everything for a migration this isn’t an issue.

We need to leave the {{promotionId}} empty first and migrate these associations in a later {{DataSet}} or always migrate “promotions” together with “orders & customers”.

### LINKED ISSUE #10610: SW6->SW6: Product `cmsPageId` prevents migration of products if set
The {{cmsPageId}} field is a foreign key on the same {{cms_page}} table and thus the referenced Id’s need to be there (which currently isn’t the case because we migrate products first, then cms pages).

We need to leave the {{cmsPageId}} empty first and migrate these associations in a later {{DataSet}} (together with the cms page {{DataSelection}})

### LINKED ISSUE #10611: SW6->SW6: Product `canonicalProductId` prevents migration of products if set
The {{canonicalProductId}} field is a foreign key on the same {{product}} table and thus the referenced Id’s need to be there (which currently isn’t always given).

We need to leave the {{canonicalProductId}} empty first and migrate these associations in a later {{DataSet}}

jozsefdamokos:
- strip late self/forward FK fields during initial migration writes
- migrate product canonicalProductId and cmsPageId in follow-up datasets
- migrate sales channel homeCmsPageId in a follow-up dataset after CMS pages
- keep hreflangDefaultDomainId on its existing delayed domain-based path and harden the first pass
- strip promotionId defensively to avoid FK failures when promotions are not migrated
- add converter fixtures/tests and admin snippet labels for the new simulated datasets

fixes: https://github.com/shopware/shopware/issues/10610 https://github.com/shopware/shopware/issues/10609 https://github.com/shopware/shopware/issues/10611


tests/_fixtures/Shopware6/Promotion/02-StripPromotionAssociation/input.php
```diff
@@ -0,0 +1,11 @@
+<?php declare(strict_types=1);
+/*
+ * (c) shopware AG <info@shopware.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+$input = require __DIR__ . '/../01-HappyCase/input.php';
```
larskemper: Are we doing this elsewhere in the fixtures? If not, I would not introduce inheritance between them.

larskemper (approved):

src/DependencyInjection/shopware6.php
```diff
@@ -1110,6 +1172,14 @@
         ])
         ->tag('shopware.migration.writer');
 
+    $services->set(SalesChannelHomeCmsPageAssociationWriter::class)
+        ->parent(AbstractWriter::class)
+        ->args([
+            service(EntityWriter::class),
+            service(\Shopware\Core\System\SalesChannel\SalesChannelDefinition::class),
```
vintagesucks:
```suggestion
            service(SalesChannelDefinition::class),
```

vintagesucks (approved):

DennisGarding (approved):

PR #194, opened 2026-06-02
fix: missing status in converted newsletter recipient

PantherX99:
Status was missing in the converted object after the last commit due to it never being set.

The status was only stored in a variable and not assigned in the converted object. This caused a warning due to a missing field "status" in the newsletter recipients migration flow.


jozsefdamokos (approved):

jozsefdamokos (commented):
Can we update `NewsletterRecipientConverterTest::testConvert` too to check status is set

jozsefdamokos (approved):

vintagesucks (approved):

MalteJanz (approved):

larskemper (approved):

PR #196, opened 2026-06-03
fix: Remove null caching for invalid language lookup

PantherX99:
Stop caching null values for languages because they might get created afterwards in the converters.

If null values are cached, converters are using inconsistent data.

Example:
1. LanguageConverter looks up language with locale ro-RO -> not found -> null gets cached for this locale
2. LanguageConverter creates this language
3. SalesChannelConverter uses the languageLookup again, but the cache still resolves to null even though a language was created by the LanguageConverter
4. Language is not assigned to SalesChannel -> Customers are getting migrated to a sales channel that do not have their language assigned -> migration fails

An different solution would be calling `$this->languageLookup->reset();` after the langauge was created in the LanguageConverter. Then all languages would need to be cached again. This has to be called everywhere, when a new language is created. I would think, having invalid locales is highly unlikely, so not caching null values would not result in performance issues.


PantherX99:
This also applies to the TranslationConverter in which translations are not being created due to a null value from the languageLookup

jozsefdamokos:
@PantherX99 thanks for this one too 👍  Can you add a test?

PantherX99:
@jozsefdamokos Like this? 
I wonder why no one had these issues before and reported them. Did they all started from scratch instead of migrating or did they not use any other language than German and English?

tests/Migration/Mapping/Lookup/LanguageLookupTest.php
```diff
@@ -58,6 +59,63 @@ public function testGetShouldGetDataFromCache(string $localeCode, ?string $expec
...
+        $locale = $connection->fetchAssociative('
+            SELECT LOWER(HEX(locale.id)) AS id, locale.code
+            FROM locale
+            LEFT JOIN language ON language.locale_id = locale.id
+            WHERE language.id IS NULL
+            LIMIT 1
+        ');
+
+        if ($locale === false) {
+            static::markTestSkipped('No locale without language found.');
```
larskemper: Tests should be deterministic. Please insert the data within the test and clean it up afterward.

tests/Migration/Mapping/Lookup/LanguageLookupTest.php
```diff
@@ -58,6 +59,63 @@ public function testGetShouldGetDataFromCache(string $localeCode, ?string $expec
...
+
+        $languageLookup = new LanguageLookup(
+            $languageRepository,
+            new LocaleLookup($localeRepository)
+        );
+
+        static::assertNull($languageLookup->get($locale['code'], $context));
+
+        $languageId = Uuid::randomHex();
+        $languageCreated = false;
```
larskemper: if `->create(...)` throws there is nothing to cleanup. So we could remove the created flag here:

```php
  $languageRepository->create([[...]], $context);
  
  try {
      static::assertSame($languageId, $languageLookup->get($locale['code'], $context));
  } finally {
      $languageRepository->delete([['id' => $languageId]], $context);
  }
```

jozsefdamokos (approved):

tests/Migration/Mapping/Lookup/LanguageLookupTest.php
```diff
@@ -58,6 +58,55 @@ public function testGetShouldGetDataFromCache(string $localeCode, ?string $expec
...
 
+    public function testGetDoesNotCacheMissingLanguage(): void
+    {
+        $context = Context::createDefaultContext();
+
+        $languageRepository = static::getContainer()->get('language.repository');
+        $localeRepository = static::getContainer()->get('locale.repository');
+
+        $localeId = Uuid::randomHex();
+        $localeCode = 'zz-ZZ';
```
vintagesucks: To fix the failing test:

```suggestion
        $localeCode = 'en-US-' . Uuid::randomHex();
```

(`zz-ZZ` is rejected by the stricter locale validation introduced in shopware/shopware#16285)

vintagesucks (approved):

PR #206, opened 2026-07-10
chore: clean up bin scripts and fix sw5 db init

larskemper:
### Changes

- Remove obsolete dev scripts: pre-commit.sh, setup.sh, fix-cs.sh, phpunit.sh, phpunit-coverage.sh, check-todos.sh, install-5-test-data.sh
- Refactor & fix source db init script
- Rename composer scripts: install5db → db:init-sw5, admin:install → npm:init (duplicate); wire npm:init into setup


jozsefdamokos (approved):

bin/install-shopware5-source-db.sh
```diff
@@ -0,0 +1,98 @@
...
+
+mysql_as() {
+    local u="$1" p="$2"; shift 2
+
+    if [[ -n "$p" ]]; then
+        mysql -u"$u" -p"$p" --host "$host" --port "$port" "$@"
+    else
+        mysql -u"$u" --host "$host" --port "$port" "$@"
+    fi
+}
```
vintagesucks, line 68: In my Docker setup this only works with `--disable-ssl-verify-server-cert` 😅

```suggestion
mysql_as() {
    local u="$1" p="$2"; shift 2
    local args=(
        --disable-ssl-verify-server-cert
        -u"$u"
        --host "$host"
        --port "$port"
    )

    if [[ -n "$p" ]]; then
        mysql "${args[@]}" -p"$p" "$@"
    else
        mysql "${args[@]}" "$@"
    fi
}
```

vintagesucks (approved):

bin/install-shopware5-source-db.sh
vintagesucks: > --ssl-mode=PREFERRED

I use `shopware/docker-dev:php8.5-node24-caddy` with plugins like `SwagMigrationAssistant` mounted into `custom/plugins`, so I have to run `db:init-sw5` inside the container for relative paths like `'../../../.env'` to work. Since the image is alpine based, `mysql` is actually `mariadb-client` and now fails with this:

→ `mysql: unknown variable 'ssl-mode=PREFERRED'` 🙃

MalteJanz (approved):

bin/install-shopware5-source-db.sh
larskemper: same for me with `--disable-ssl-verify-server-cert` 😄 will look for something diff

larskemper: i added a mariadb flag. So `composer run db:init-sw5 -- --mariadb` should work now

vintagesucks: Works! I'll use `MIGRATION_DB_CLIENT=mariadb` + `composer setup`, which works too 👍

PR #215, opened 2026-07-30
chore: Prepare code for PHPStan update 2026-07

mitelg:
Preparation PR for https://github.com/shopware/shopware/pull/17661

### DIFF
--- a/src/Migration/Mapping/MappingService.php
+++ b/src/Migration/Mapping/MappingService.php
@@ -126,6 +126,7 @@ class MappingService implements MappingServiceInterface, ResetInterface
                 WHERE connection_id = :connectionId
                     AND entity = :entity
                     AND old_identifier = :oldIdentifier;';
+        /** @var array{id: string, connectionId: string, entity: string|null, oldIdentifier: string|null, entityId: string|null, entityValue: string|null, checksum: string|null, additionalData: string|null}|false $mapping */
         $mapping = $this->connection->fetchAssociative($sql, ['connectionId' => Uuid::fromHexToBytes($connectionId), 'entity' => $entityName, 'oldIdentifier' => $oldIdentifier]);
         if ($mapping === false) {
             return null;
--- a/src/Profile/Shopware/Converter/MainVariantRelationConverter.php
+++ b/src/Profile/Shopware/Converter/MainVariantRelationConverter.php
@@ -58,7 +58,7 @@ abstract class MainVariantRelationConverter extends ShopwareConverter
 
         if (
             $mainProductMapping === null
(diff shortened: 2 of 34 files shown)


larskemper (approved):

src/Migration/Mapping/MappingService.php
```diff
@@ -126,6 +126,7 @@ public function getMapping(
                 WHERE connection_id = :connectionId
                     AND entity = :entity
                     AND old_identifier = :oldIdentifier;';
+        /** @var array{id: string, connectionId: string, entity: string|null, oldIdentifier: string|null, entityId: string|null, entityValue: string|null, checksum: string|null, additionalData: string|null}|false $mapping */
```
jozsefdamokos, line 129: can we not use MappingStruct?

src/Profile/Shopware/Converter/ShippingMethodConverter.php
```diff
@@ -300,7 +300,7 @@ protected function addShippingMethodTranslation(array &$shippingMethod, array $d
     /**
      * @param array<string, mixed> $data
      *
-     * @return array<string, mixed>
+     * @return Rules
```
jozsefdamokos: Should we rename this? the method returns a single rule. Also `Rules` seems a bit too generic

src/Profile/Shopware/Converter/MainVariantRelationConverter.php
```diff
@@ -58,7 +58,7 @@ public function convert(array $data, Context $context, MigrationContextInterface
 
         if (
             $mainProductMapping === null
-            || !\is_string($mainProductMapping['entityId'] ?? null)
+            || !\is_string($mainProductMapping['entityId'])
```
jozsefdamokos, line 61: Is `entityId` always set?

jozsefdamokos (approved):

src/Profile/Shopware/Converter/ShippingMethodConverter.php
mitelg: done! :+1:

src/Migration/Mapping/MappingService.php
mitelg: no, it is almost the same expect the `additionalData` which is just a encoded JSON string, when read from the DB :wink:

src/Profile/Shopware/Converter/MainVariantRelationConverter.php
mitelg: it is indeed and it is either a string or null, so checking for `null` and then set to `null` has no effect

PR #216, opened 2026-08-07
fix: terminate failed media downloads after retry threshold

### LINKED ISSUE #18902: [Migration][Magento] Unreachable media source: Downloading step does not converge, migration cannot reach Done
### Shopware Version

6.7.2.2

### Affected area / extension

Extension:Migration-Tool

### Actual behaviour

Plugin: SwagMigrationMagento 13.0.0 on Migration Assistant 18.x, local gateway.

When media is fetched over HTTP and the source is unreachable, a failed fetch is not marked processed
and not marked failed — it is retried. Each retry re-increments the Downloading progress counter, so
progress goes past the file total (observed 154%–164% on one run), the step does not reach a terminal
state, swag_migration_logging keeps growing, and the migration cannot reach Done. The only exit found
was a manual DB update. The retry is bounded in code by migrationDefaultExceptionThreshold, but a file
is flagged as failure only once its per-file errorCount exceeds that threshold; over ~6,000 unreachable
files the step did not converge in the observation window (every row stayed processed=0, process_failure=0).

Code (line numbers as of plugin v13.0.0; present on trunk):
- src/Profile/Magento/Media/LocalMediaProcessor.php, LocalMediaProcessor::downloadMediaFiles(), line 437:
  on a non-'fulfilled' state the uuid gets setErrorCount(+1) and is only added to $failureUuids /
  ERROR_STATE when getErrorCount() > migrationDefaultExceptionThreshold, otherwise `continue` — so
  setProcessedFlag() (end of the method) marks it neither processed nor process_failure, and it is
  re-enqueued.

Note: the "progress exceeds 100%" symptom is a measured observation from the run; the progress-accounting
path lives in the parent Migration Assistant and was not traced here.

### Expected behaviour

An unreachable media file should reach a terminal state (failed / skipped) in a bounded, small number of
attempts, the progress counter should not exceed the file total, and the step should complete so the
migration can reach Done. An operator should not need direct DB access to finish a migration because some
images could not be downloaded. Ideally, probe the media source's reachability up front and warn before
the run.

### How to reproduce

1. Create a migration connection whose media installationRoot is not reachable from the Shopware
   container (e.g. import only the DB, or point at a host the container cannot resolve).
2. Run a migration that pulls media (product images come with Products).
3. Watch the Downloading step: progress climbs past the file total (e.g. 9,180 -> 9,830 against a total of
   5,970) and does not settle; swag_migration_media_file rows stay written=1, processed=0, process_failure=0.
4. The step never reaches a terminal state on its own; Done is unreachable.
Manual unblock used in the lab:
   UPDATE swag_migration_media_file SET processed = 1, process_failure = 1 WHERE processed = 0;

DennisGarding:
Closes:  https://github.com/shopware/shopware/issues/18902

When a media file cannot be fetched from Magento, the Migration Assistant retries the file without marking it as processed or failed. This causes the downloading progress to exceed the total number of files and prevents the migration from reaching a terminal state.

Failed media downloads are marked as failed once the configured retry threshold is reached. This ensures that:
- each media file reaches a terminal state;
- failed files are not re-enqueued indefinitely;
- the download progress cannot grow beyond the total file count;
- migrations can complete even when individual media sources are unreachable.

Successful downloads keep their existing processing behaviour.


src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
```diff
@@ -220,18 +227,25 @@ private function getMediaFiles(MigrationContextInterface $migrationContext): arr
...
+            $mappedWorkload[$item->getMediaId()] = $item;
+        }
+
         for ($i = 0; $i < $this->migrationConfig->migrationDefaultExceptionThreshold; ++$i) {
             $errorWorkload = [];
 
-            foreach ($workload as $item) {
-                if ($item->getErrorCount() > 0) {
+            foreach ($mappedWorkload as $item) {
+                if ($item->getErrorCount() > 0 && $item->getState() !== MediaProcessWorkloadStruct::ERROR_STATE) {
```
larskemper, line 250: `HttpDownloadServiceBase` does not always set `ERROR_STATE` just like Magento: https://github.com/shopware/SwagMigrationAssistant/blob/trunk/src/Migration/Media/Processor/HttpDownloadServiceBase.php#L295

src/Migration/MessageQueue/Handler/Processor/MediaProcessingProcessor.php
```diff
@@ -179,7 +179,14 @@ public function process(
             );
         }
 
-        $workloadCount = \count($workload);
+        $workloadCount = \count(\array_filter(
```
larskemper, line 184: Isn't there a possibility now that the migration run gets stale? Like when for some reason the fetch fails and the media is stuck in progress the process would never get to 100%, cause we filter them out

src/Migration/Media/Processor/HttpDownloadServiceBase.php
```diff
@@ -291,8 +300,7 @@ private function doNormalDownloadRequest(MigrationContextInterface $migrationCon
                     ->build(RunExceptionLog::class)
             );
 
-            $promise = null;
-            $workload->setErrorCount($workload->getErrorCount() + 1);
```
jozsefdamokos, line 295: Dont we still need to increase the error count here?

src/Profile/Shopware6/Converter/LanguageConverter.php
```diff
@@ -66,7 +66,9 @@ protected function convertData(array $data): ConvertStruct
             );
         }
 
-        return new ConvertStruct($converted, $data, $this->mainMapping['id'] ?? null);
+        $mappingId = isset($this->mainMapping['id']) ? $this->mainMapping['id'] : null;
```
jozsefdamokos, line 69: Isnt this the same?

phpstan.neon.dist
```diff
@@ -21,6 +21,12 @@ parameters:
         narrow_param: false
 
     ignoreErrors:
+        # The builder instantiates different concrete log entries from a validated class-string.
+        # The generic @return T annotation already provides the precise type at call sites.
+        - identifier: typePerfect.narrowReturnObjectType
```
jozsefdamokos, line 26: Why is this needed now? You didnt touch `MigrationLogBuilder` 🤔

jozsefdamokos (approved):

phpstan.neon.dist
DennisGarding: I dont know why the pipeline is failing without it. Maybe Core update that now takes effect.

src/Migration/Media/Processor/HttpDownloadServiceBase.php
DennisGarding: The error count is still increased, but later in promise result handled. doNormalDownloadRequest() returns a rejected promise, and the failure branch in process() increments the count exactly once.

src/Profile/Shopware6/Converter/LanguageConverter.php
DennisGarding: Yes, you are right. I revert that

larskemper: fixed with https://github.com/shopware/SwagMigrationAssistant/pull/217

phpstan.neon.dist
larskemper: fixed with https://github.com/shopware/SwagMigrationAssistant/pull/217

src/Profile/Shopware6/Converter/LanguageConverter.php
DennisGarding: ok, now i remember. Its a phpstan issue:

```
 ------ --------------------------------------------------------------------- 
  Line   src/Profile/Shopware6/Converter/LanguageConverter.php                
 ------ --------------------------------------------------------------------- 
  69     Coalesce operator ?? is unnecessary because the left side is always  
         set and the right side is null.                                      
         🪪  nullCoalesce.unnecessary                                         
 ------ --------------------------------------------------------------------- 
```

Wait for https://github.com/shopware/SwagMigrationAssistant/pull/217 is merged

phpstan.neon.dist
vintagesucks: reopening, this still needs to be reverted, no?

src/Migration/Media/Processor/HttpDownloadServiceBase.php
vintagesucks, line 1: The changes in this PR could use a `CHANGELOG` entry

vintagesucks (approved):

phpstan.neon.dist
DennisGarding: No, as you can see, #217 is merged and its not fixed. Maybe a missunderstood?!

larskemper: It clearly ignores the same thing. We should go with one of the options. I choose the in-code ignore cause it will stay and its just a single fail. Please try removing the global ignore and let the ci run again

DennisGarding: ah ok

PR #218, opened 2026-08-14
ci: verify extension against lowest dependencies

vintagesucks:
follow-up to shopware/SwagMigrationMagento#62 

- `AttributeReader::cleanupColumns()` used `getReferencingColumnNames()` and `UnqualifiedName` (DBAL 4.3+) since #67, so attribute reads through the Shopware local gateway failed on installs whose lockfiles contain an older DBAL
- the new `Extension Verifier (lowest)` job uses `shopware-cli` via [`shopware/github-actions/extension-verifier`](https://github.com/shopware/github-actions/tree/main/extension-verifier) to [validate](https://developer.shopware.com/docs/products/tools/cli/validation.html) the plugin against its lowest resolvable dependencies to catch issues like this in CI in the future
- `shopware/storefront` added to `suggest`, the verifier installs suggested packages and needs them to resolve the theme classes `RunService` pulls in
- the two Symfony DI configurator findings ignored in `.shopware-extension.yml` instead of fixed in code, the only in-code fix would be moving `src/DependencyInjection/*.php` to `src/Resources/config/` which felt too heavy here

example output from the [first run of the job](https://github.com/shopware/SwagMigrationAssistant/actions/runs/31796290825/job/94754024779?pr=218):

```text
phpstan.neon
  0  error    [phpstan/error]  Internal error: Class Doctrine\DBAL\Schema\Name\UnqualifiedName was not found while trying to analyse it - discovering symbols is probably not configured properly. while analysing file /tmp/analyse-extension-1320630473/src/Profile/Shopware/Gateway/Local/Reader/AttributeReader.php

✖ 10 problems (5 errors, 5 warnings)
```

### DIFF
--- a/.github/workflows/php.yaml
+++ b/.github/workflows/php.yaml
@@ -40,7 +40,7 @@ jobs:
               run: composer config policy.advisories.block false
 
             - name: Run verifier
-              uses: shopware/github-actions/extension-verifier@extension-verifier-exclude
+              uses: shopware/github-actions/extension-verifier@main
               with:
                   action: check
                   check-against: lowest
--- a/.github/workflows/php.yaml
+++ b/.github/workflows/php.yaml
@@ -44,3 +44,4 @@ jobs:
               with:
                   action: check
                   check-against: lowest
+                  exclude: stylelint
--- a/.github/workflows/php.yaml
+++ b/.github/workflows/php.yaml
@@ -40,7 +40,7 @@ jobs:
               run: composer config policy.advisories.block false
 
             - name: Run verifier
-              uses: shopware/github-actions/extension-verifier@main
+              uses: shopware/github-actions/extension-verifier@extension-verifier-exclude
               with:
                   action: check
                   check-against: lowest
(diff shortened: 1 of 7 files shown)


.github/workflows/php.yaml
```diff
@@ -25,3 +25,23 @@ jobs:
...
+              uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+            - name: Remove PHPStan config
+              run: rm phpstan.neon.dist
+
+            - name: Allow composer to resolve advisory-affected releases
+              run: composer config policy.advisories.block false
+
+            - name: Run verifier
+              uses: shopware/github-actions/extension-verifier@extension-verifier-exclude
```
vintagesucks:
```suggestion
              uses: shopware/github-actions/extension-verifier@main
```

needs https://github.com/shopware/github-actions/pull/192

larskemper (approved):

MalteJanz (approved):

PR #221, opened 2026-09-07
fix: phpstan playwright build

Pierre02:

1. PHPStan: Deprecation not yet resolved were causing the pipeline to fail
2. There was a complete design change in the admin which broke:

- Screenshot expectation
- Selector to truncate connector, leading to failing tests down the line since it wasn't deleted

// closes https://github.com/shopware/shopware-private/issues/408

- Deprecations with specific tags and string are excluded from PHPStan validation
- Selector was fixed to properly enable connector deletion
- Unnecessary page reload waiter were removed to avoid false fail in the pipeline
- Screenshot were regenerated to reflect the whole new design

### 3. Describe each step to reproduce the issue or behaviour.

- Run the nightly build on the trunk

### 5. Checklist

- [x] I have created the PR in draft status and only open it when it's ready for review
- [x] If the change is large: I have thought about splitting it up into smaller PRs
- [x] I have written tests and if it's a bug fix verified that they fail without my change or that new code is covered by tests
- [x] I have updated developer-facing release notes if this change affects external developers
- [x] I have written or adjusted the documentation according to my changes
- [x] I added the correct package annotation to each `src` file (not strictly necessary for tests)
- [x] This change has code comments where appropriate, especially for non-obvious lines of code
- [x] Ensure the pull request title as well as first commit follows conventional commits
- [x] I have thought about well-defined extension points and marked everything else as `@internal` / `@private`


larskemper (approved):
first pr 🎉 lgtm

jozsefdamokos (approved):

PR #222, opened 2026-09-08
fix: phpstan playwright build

Pierre02:

The nightly was still not running due to changes in core for jest tests

Uses DomWrapper where appropriate to reflect core changes

### 3. Describe each step to reproduce the issue or behaviour.

Run the jest tests

### 5. Checklist

- [x] I have created the PR in draft status and only open it when it's ready for review
- [x] If the change is large: I have thought about splitting it up into smaller PRs
- [x] I have written tests and if it's a bug fix verified that they fail without my change or that new code is covered by tests
- [x] I have updated developer-facing release notes if this change affects external developers
- [x] I have written or adjusted the documentation according to my changes
- [x] I added the correct package annotation to each `src` file (not strictly necessary for tests)
- [x] This change has code comments where appropriate, especially for non-obvious lines of code
- [x] Ensure the pull request title as well as first commit follows conventional commits
- [x] I have thought about well-defined extension points and marked everything else as `@internal` / `@private`


tests/Jest/src/module/swag-migration/component/swag-migration-error-resolution/swag-migration-error-resolution-modal.spec.js
vintagesucks: (nit) I think this one needs the new body scope too, the `exists` check currently can't fail

phpstan.neon.dist
vintagesucks, line 64: (nit) The PR currently also contains the changes from #221 (you probably reused the branch), which makes it a bit confusing to review 👀

vintagesucks (approved):

phpstan.neon.dist
Pierre02: Yes I was also surprised, since the changes are already in trunk I didn't expect them to show up here.

vintagesucks: #221 was squash-merged into trunk, meaning the 3 commits you had locally technically aren't on trunk, so you re-using your local branch, now behind a trunk only containing [one squashed commit](https://github.com/shopware/SwagMigrationAssistant/commit/670ea2b8c3e16f675c5c6825fb802c1576146d11), included them again.

Pierre02: Ah thank you, I missed the squashed part... Makes sense, and I will make sure it doesn't happen again!

vintagesucks (approved):

DennisGarding (approved):

